PR #2, opened 2015-01-20
Implement admin leave

jwolski:
Excuse all the churn here. 

The operation implemented is an /admin/leave which sets the local member status to leave and then disseminated to the protocol pingers. This operation is manual much like /admin/join. When a member is in the 'leave' status, the member is soft-removed from membership. It'll be removed from pingers lists and the ring itself. As part of the 'leave' operation, local gossip and suspect protocol are stopped. A member that has previously left can rejoin by /admin/join.

As part of this work I refactored and took the start/stop of gossip and suspicion subprotocol out into the lib/swim.js file. I think I left this code slightly better than it was. Still lots of room for improvement.

There's also a relatively high amount of test coverage here... which is good

@mranney @Raynos

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,6 +28,7 @@ var TypedError = require('error/typed');
 var AdminJoiner = require('./lib/swim').AdminJoiner;
 var createRingPopTChannel = require('./lib/tchannel.js').createRingPopTChannel;
 var Dissemination = require('./lib/members').Dissemination;
+var Gossip = require('./lib/swim.js').Gossip;
 var HashRing = require('./lib/ring');
 var Membership = require('./lib/members').Membership;
 var MemberIterator = require('./lib/members').MemberIterator;
@@ -36,9 +37,25 @@ var PingReqSender = require('./lib/swim').PingReqSender;
 var PingSender = require('./lib/swim').PingSender;
 var safeParse = require('./lib/util').safeParse;
 var RequestProxy = require('./lib/request-proxy');
+var Suspicion = require('./lib/swim.js').Suspicion;
 
 var HOST_PORT_PATTERN = /^(\d+.\d+.\d+.\d+):\d+$/;
 var MAX_JOIN_DURATION = 300000;
+
+var AppRequiredError = TypedError({
+    type: 'ringpop.options-app.required',
+    message: 'Expected `options.app` to be a non-empty string.\n' +
+        'Must specify an app for ringpop to work.\n'
+});
+
+var HostPortRequiredError = TypedError({
+    type: 'ringpop.options-host-port.required',
+    message: 'Expected `options.hostPort` to be valid.\n' +
+        'Got {hostPort} which is not {reason}.\n' +
+        'Must specify a HOST:PORT string.\n',
+    hostPort: null,
+    reason: null
+});
 
 var InvalidJoinAppError = TypedError({
     type: 'ringpop.invalid-join.app',
@@ -55,10 +72,9 @@ var InvalidJoinSourceError = TypedError({
     actual: null
 });
 
-var AppRequiredError = TypedError({
-    type: 'ringpop.options-app.required',
-    message: 'Expected `options.app` to be a non-empty string.\n' +
-        'Must specify an app for ringpop to work.\n'
+var InvalidLocalMemberError = TypedError({
+    type: 'ringpop.invalid-local-member',
+    message: 'Operation could not be performed because local member has not been added to membership'
 });
 
 var OptionsRequiredError = TypedError({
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -109,7 +119,8 @@ require('util').inherits(RingPop, EventEmitter);
 
 RingPop.prototype.destroy = function destroy() {
     this.destroyed = true;
-    clearTimeout(this.gossipTimer);
+    this.gossip.stop();
```
Raynos: nitpick: favor having a method called `.destroy()`

Every single thing that has a side effect in terms of keeping the event loop open in node core has an `unref()` or `destroy()` method.

With the exception that servers have a `close()` method.

index.js
```diff
@@ -140,7 +146,21 @@ RingPop.prototype.setupChannel = function setupChannel() {
     createRingPopTChannel(this, this.channel);
 };
 
+RingPop.prototype.addLocalMember = function addLocalMember() {
+    this.membership.addMember({ address: this.hostPort });
+};
+
 RingPop.prototype.adminJoin = function adminJoin(target, callback) {
+    if (!this.membership.localMember) {
+        return callback(InvalidLocalMemberError());
```
Raynos: nitpick: For sanity a callback must always be EITHER sync or async. 

See http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony

Recommend doing `process.nextTick(flushError)`

lib/swim.js
```diff
@@ -250,8 +252,169 @@ PingSender.prototype.doCallback = function doCallback(isOk, bodyObj) {
     }
 };
 
+function Gossip(ringpop) {
+    this.ringpop = ringpop;
+    this.isStopped = null;
```
Raynos: nitpick: consider defaulting to `true`

lib/swim.js
```diff
@@ -250,8 +252,169 @@ PingSender.prototype.doCallback = function doCallback(isOk, bodyObj) {
...
+    this.isStopped = true;
+
+    this.ringpop.logger.debug('stopped gossip protocol', {
+        local: this.ringpop.membership.getLocalMemberAddress()
+    });
+};
+
+Gossip.prototype.restart = function restart() {
+    if (this.isStopped !== true) {
+        this.ringpop.logger.warn('cannot restart gossip protocol because it was never stopped', {
```
Raynos: A restart method makes sense in both `STOP` and `START` state.

It should be `if (started) { stop() }; start();`

Otherwise this mehtod might as well be called `.start()`

test/mock/index.js
```diff
@@ -17,9 +17,12 @@
...
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
+var path = require('path');
 
-module.exports = {
-    channel: require('./channel.js'),
-    logger: require('./logger.js'),
-    membership: require('./membership.js')
-};
+require('glob').sync(__dirname + '/**/*.js').forEach(function(file) {
```
Raynos: nitpick: This style of loading modules makes it difficult to know what it actually is.

Having an object literal makes reading the code very easy, you just read the code.

Seeing a glob and a bunch of meta programming means you have to `ls __dirname + **/*.js` and then figure out what it actually is.

test/mock/noop.js
```diff
@@ -0,0 +1 @@
+module.exports = function() {};
```
Raynos: `npm i noop2` ( https://github.com/yoshuawuyts/noop2 ) :D 

Don't actually do that ...

lib/members.js
```diff
@@ -23,6 +23,7 @@ var EventEmitter = require('events').EventEmitter;
 var util = require('util');
 
 var LOG_10 = Math.log(10);
+var MEMBER_STATUSES = ['alive', 'faulty', 'leave', 'suspect'];
```
Raynos: nitpick: make this an object whose value is always true.

i.e. `{ alive: true, faulty: true, ... }`

Then you can do `MEMBER_STATUSES[status]` instead of `indexOf()`

Raynos:
@jwolski This lgtm.

I don't know enough about the SWIM protocol and the gossip to very distributed correctness.

test/mock/noop.js
jwolski: :)

index.js
jwolski: i don't like `destroy()` because it has a sense of finality to it.in this case, with the gossip object, a `stop()` can be `start()`'ed again. calling destroy() and then start() again would be weird.

Raynos: Makes sense. :+1:

index.js
jwolski: thanks. helpful feedback.

lib/swim.js
jwolski: yea, this is awkward. will fix.

test/mock/index.js
jwolski: (okay)

PR #3, opened 2015-01-22
Move some packages into dev deps

jwolski:
@Raynos

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "test": "npm run jshint && node test/index.js"
   },
   "dependencies": {
-    "async": "^0.9.0",
     "body": "^4.5.0",
     "error": "^5.0.0",
     "farmhash": "^0.2.0",
--- a/package.json
+++ b/package.json
@@ -19,19 +19,17 @@
   "dependencies": {
     "async": "^0.9.0",
     "body": "^4.5.0",
-    "cli-color": "^0.3.2",
-    "commander": "^2.6.0",
-    "debuglog": "^1.0.1",
     "error": "^5.0.0",
-    "express": "^4.9.5",
     "farmhash": "^0.2.0",
     "metrics": "^0.1.8",
     "readable-stream": "^1.0.33",
-    "time-mock": "^0.1.2",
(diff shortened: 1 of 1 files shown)


Raynos:
lgtm.

package.json
```diff
@@ -19,19 +19,17 @@
   "dependencies": {
     "async": "^0.9.0",
```
Raynos: Get rid of async, no longer used.

package.json
```diff
@@ -19,19 +19,17 @@
...
-    "cli-color": "^0.3.2",
-    "commander": "^2.6.0",
-    "debuglog": "^1.0.1",
     "error": "^5.0.0",
-    "express": "^4.9.5",
     "farmhash": "^0.2.0",
     "metrics": "^0.1.8",
     "readable-stream": "^1.0.33",
-    "time-mock": "^0.1.2",
     "underscore": "^1.5.2"
```
Raynos: We can refactor this dependency away.

package.json
```diff
@@ -19,19 +19,17 @@
   "dependencies": {
     "async": "^0.9.0",
     "body": "^4.5.0",
```
Raynos: we can get this dependency replace with something lighter or something inline.

Raynos: I should probably update it to use raw-body instead.

package.json
```diff
@@ -19,19 +19,17 @@
...
-    "cli-color": "^0.3.2",
-    "commander": "^2.6.0",
-    "debuglog": "^1.0.1",
     "error": "^5.0.0",
-    "express": "^4.9.5",
     "farmhash": "^0.2.0",
     "metrics": "^0.1.8",
     "readable-stream": "^1.0.33",
-    "time-mock": "^0.1.2",
```
Raynos: Is this not being used in tests? Shouldnt this be a dev dependency ?

jwolski: it's already listed as a `devDependency` so it was redundant in that sense

package.json
jwolski: soon

PR #8, opened 2015-01-22
gaurd against invalid app

Raynos:
reviewers: @jwolski

cc: @sh1mmer

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,7 +37,6 @@ var PingSender = require('./lib/swim').PingSender;
 var safeParse = require('./lib/util').safeParse;
 var RequestProxy = require('./lib/request-proxy');
 
-var IP_PATTERN = /^(\d+.\d+.\d+.\d+)$/;
 var HOST_PORT_PATTERN = /^(\d+.\d+.\d+.\d+):\d+$/;
 var MAX_JOIN_DURATION = 300000;
 
@@ -95,16 +94,14 @@ function RingPop(options) {
     var isString = typeof options.hostPort === 'string';
     var parts = options.hostPort && options.hostPort.split(':');
     var isColonSeparated = parts && parts.length === 2;
-    var isIP = parts && parts[0] && IP_PATTERN.test(parts[0]);
     var isPort = parts && parts[1] &&
         !isNaN(parseInt(parts[1], 10));
 
-    if (!isString || !isColonSeparated || !isIP || !isPort) {
+    if (!isString || !isColonSeparated || !isPort) {
         throw HostPortRequiredError({
             hostPort: options.hostPort,
             reason: !isString ? 'a string' :
                 !isColonSeparated ? 'a valid hostPort pattern' :
-                !isIP ? 'a valid ip' :
                 !isPort ? 'a valid port' : 'correct'
         });
     }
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -53,11 +53,23 @@ var InvalidJoinSourceError = TypedError({
...
+    message: 'Expected `options.app` to be a non-empty string.\n' +
+        'Must specify an app for ringpop to work.\n'
+});
+
 function RingPop(options) {
     if (!(this instanceof RingPop)) {
         return new RingPop(options);
     }
 
+    if (typeof options.app !== 'string' ||
```
jwolski: do `hostPort` too?

Raynos: fixed.

Raynos:
@jwolski Updated this branch with more checks. ALL THE CHECKS.

index.js
```diff
@@ -53,11 +56,59 @@ var InvalidJoinSourceError = TypedError({
...
+    hostPort: null,
+    reason: null
+});
+
 function RingPop(options) {
     if (!(this instanceof RingPop)) {
         return new RingPop(options);
     }
 
+    if (!options) {
```
jwolski: SO MUCH CODEZ.

_required_ options always rubs me the wrong way. It's a contradiction. Rename to `params` now that we're strictly enforcing. Thoughts?

index.js
```diff
@@ -53,11 +56,59 @@ var InvalidJoinSourceError = TypedError({
...
+
+    if (typeof options.app !== 'string' ||
+        options.app.length === 0
+    ) {
+        throw AppRequiredError();
+    }
+
+    var isString = typeof options.hostPort === 'string';
+    var parts = options.hostPort && options.hostPort.split(':');
+    var isColonSeperated = parts && parts.length === 2;
```
jwolski: spelling: separated

index.js
```diff
@@ -53,11 +56,59 @@ var InvalidJoinSourceError = TypedError({
...
+    var isIP = parts && parts[0] && parts[0].match(IP_PATTERN);
```
jwolski: do you need to `match`? How about using just `test()`?

index.js
Raynos: `params` is incredibly not idiomatic.

required options are normal in most npm libraries. I understand your feelings but we should stick with `opts` or `options`.

Raynos: > SO MUCH CODEZ

Good user experience requires so much validation boilerplate ;_;

index.js
jwolski: asserting on an IP is going to fuck people up. current apps use hostname for host:port

index.js
jwolski: ill lean on you for what's more conventional. If you say stick with `options`, let's stick with `options`.

index.js
Raynos: @jwolski do we allow hosts ? we assert on IPs in other places.

jwolski: yes, we allow hosts.

Raynos:
@jwolski addressed review feedback.

jwolski:
LGTM

Raynos:
Published v4.1.0

PR #12, opened 2015-01-27
Add key used on lookup to head transmitted through proxy and emit head too

jwolski:
This is a breaking change required for serial processing

@Raynos

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,6 +19,7 @@
 // THE SOFTWARE.
 'use strict';
 
+var _ = require('underscore');
 var EventEmitter = require('events').EventEmitter;
 var fs = require('fs');
 var metrics = require('metrics');
@@ -40,6 +41,7 @@ var Suspicion = require('./lib/swim.js').Suspicion;
 
 var HOST_PORT_PATTERN = /^(\d+.\d+.\d+.\d+):\d+$/;
 var MAX_JOIN_DURATION = 300000;
+var PROXY_REQ_PROPS = ['key', 'dest', 'req', 'res'];
 
 var AppRequiredError = TypedError({
     type: 'ringpop.options-app.required',
@@ -79,7 +81,14 @@ var InvalidLocalMemberError = TypedError({
 var OptionsRequiredError = TypedError({
     type: 'ringpop.options.required',
     message: 'Expected `options` argument to be passed.\n' +
-        'Must specify options for `RingPop({ ... })`.\n'
+        'Must specify options for `{method}`.\n',
+    method: null
+});
+
+var PropertyRequiredError = TypedError({
+    type: 'ringpop.options.property-required',
+    message: 'Expected `{property}` to be defined within options argument.',
+    property: null
 });
 
 var RedundantLeaveError = TypedError({
@@ -93,7 +102,7 @@ function RingPop(options) {
     }
 
     if (!options) {
-        throw OptionsRequiredError();
+        throw OptionsRequiredError({ method: 'RingPop' });
     }
 
     if (typeof options.app !== 'string' ||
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -77,9 +78,16 @@ var InvalidLocalMemberError = TypedError({
 });
 
 var OptionsRequiredError = TypedError({
-    type: 'ringpop.options.required',
+    type: 'ringpop.options-required',
```
Raynos: these type strings cannot be changed. sorry.

jwolski: DDDD:

lib/request-proxy.js
```diff
@@ -44,15 +44,21 @@ function RequestProxy(ringpop) {
 
 var proto = RequestProxy.prototype;
 
-proto.proxyReq = function proxyReq(dest, req, res, opts) {
+proto.proxyReq = function proxyReq(opts) {
+    var key = opts.key;
+    var dest = opts.dest;
+    var req = opts.req;
+    var res = opts.res;
+    var opts2 = opts.opts;
```
Raynos: this is weird.

Raynos: maybe put all the other opts keys flatly onto opts ?

jwolski: yes.

index.js
```diff
@@ -807,8 +826,24 @@ RingPop.prototype.handleOrProxy =
         if (this.whoami() === dest) {
             return true;
         } else {
-            this.proxyReq(dest, req, res, opts);
+            this.proxyReq({
```
Raynos: could do `extend(opts, { key, dest, req, res })`

index.js
```diff
@@ -19,6 +19,7 @@
 // THE SOFTWARE.
 'use strict';
 
+var _ = require('underscore');
```
Raynos: nitpick: `var extend = require('xtend')`

index.js
```diff
@@ -807,8 +821,23 @@ RingPop.prototype.handleOrProxy =
         if (this.whoami() === dest) {
             return true;
         } else {
-            this.proxyReq(dest, req, res, opts);
+            this.proxyReq(_.extend(opts, {
```
Raynos: would advice against a mutable extension. This might cause suprises for the caller.

lib/request-proxy.js
```diff
@@ -70,7 +75,10 @@ proto.proxyReq = function proxyReq(dest, req, res, opts) {
             headers: headers,
             method: method,
             httpVersion: httpVersion,
-            checksum: ringpop.membership.checksum
+            checksum: ringpop.membership.checksum,
+            ringpopKey: key,
+            ringpopSource: ringpop.whoami(),
+            ringpopDest: dest
```
Raynos: dest is not needed as it's available as `hostInfo` on the remote side.

Let's not send it unless you want to do packet inspection or something.

lib/request-proxy.js
```diff
@@ -70,7 +75,10 @@ proto.proxyReq = function proxyReq(dest, req, res, opts) {
             headers: headers,
             method: method,
             httpVersion: httpVersion,
-            checksum: ringpop.membership.checksum
+            checksum: ringpop.membership.checksum,
+            ringpopKey: key,
+            ringpopSource: ringpop.whoami(),
```
Raynos: source is also duplicate information that is available as `hostInfo`

Raynos:
Let's fix:

 - do not mutate arguments.

Otherwise, lgtm.

index.js
jwolski: yes, definitely. for some reason i thought it wouldn't mutate `opts` and would generate a copy.

Raynos: `xtend` generates a copy by default ( sane defaults ). `_.extend` mutates by default.

index.js
jwolski: no

lib/request-proxy.js
jwolski: we will eventually do tchannel snooping with Matt's tchannel's packet capturer. ill leave it out for now until we need it.

PR #13, opened 2015-01-27
Travis CI integration

jwolski:
@Raynos @robskillington

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: node_js
+node_js:
+  - "0.10"
+before_install: npm i npm@latest -g
+script: npm run travis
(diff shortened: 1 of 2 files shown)


robskillington:
Might you want:

```
notifications:
  email: false
```
?  The emails are quite the spam.  As per https://github.com/uber/jetstream/blob/master/.travis.yml

Also you might need to flip a switch somewhere, it should've built this PR and posted back so something might not be configured just yet.

jwolski:
thx

Raynos:
@jwolski 

See:

 - https://github.com/uber/uber-ngen/blob/master/templates/uber/content/.travis.yml
 - https://github.com/uber/uber-ngen/blob/master/templates/uber/content/package.json#L40

We want to apply these. Especially the coverage stuff for integrating with coveralls.

Raynos:
lgtm.

PR #14, opened 2015-01-30
Add a bunch of proxying logging.

Raynos:
reviewer: @jwolski

cc @sh1mmer


Raynos:
Got a verbal lgtm from @jwolski

mranney:
Note that trace logging in this way has a performance cost, especially in a hot path. We should make sure that this doesn't have a significant impact on performance in the common case.

Raynos:
@mranney agreed.

I think since `proxyReq()` is a once per request / response pair operation it's not enough of a hot path that allocating these little objects is going to become a problem.

However we might have to re-consider in the future.

PR #15, opened 2015-01-31
Faulty overrides alive member with same incarnation number

jwolski:
While unlikely that suspect hasn't made its way around the cluster, given the 5s suspect period timeout, this will help fault detection in the event that a faulty member has been detected before suspect status has reached all nodes.

@mranney @mrhooray @Raynos


Raynos:
Lgtm.

mrhooray:
lgtm.

PR #16, opened 2015-02-02
handleOrProxy opts arg is not required

jwolski:
@Raynos

### DIFF
--- a/index.js
+++ b/index.js
@@ -832,12 +832,12 @@ RingPop.prototype.handleOrProxy =
                 key: key,
                 url: req && req.url
             });
-            this.proxyReq(_.extend(opts, {
+            this.proxyReq(_.defaults({
                 key: key,
                 dest: dest,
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -832,12 +832,12 @@ RingPop.prototype.handleOrProxy =
                 key: key,
                 url: req && req.url
             });
-            this.proxyReq(_.extend(opts, {
+            this.proxyReq(_.extend({
```
Raynos: this is not the semantics you want.

You wanted `_.extend(opts || {}, { ... })`

If you set `{ key: 'foo' }` then other stuff will break horribly downstream.

Raynos:
Recommended: minor

index.js
jwolski: true

Raynos:
legit. lgtm.

PR #17, opened 2015-02-03
Use body caching

jwolski:
@Raynos

### DIFF
--- a/lib/request-proxy.js
+++ b/lib/request-proxy.js
@@ -59,7 +59,7 @@ proto.proxyReq = function proxyReq(opts) {
     var timeout = opts.timeout ?
         opts.timeout : ringpop.proxyReqTimeout;
 
-    body(req, onBody);
+    body(req, res, { cache: true }, onBody);
 
     function onBody(err, rawBody) {
         if (err) {
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "view-cover": "opn coverage/index.html"
   },
   "dependencies": {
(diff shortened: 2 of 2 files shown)


Raynos:
lgtm.

mranney:
What does body caching do?

lib/request-proxy.js
```diff
@@ -59,7 +59,7 @@ proto.proxyReq = function proxyReq(opts) {
     var timeout = opts.timeout ?
         opts.timeout : ringpop.proxyReqTimeout;
 
-    body(req, onBody);
+    body(req, res, { cache: true }, onBody);
 
     function onBody(err, rawBody) {
```
mranney: It would be faster and less GC if you moved onBody back to the prototype here.

jwolski:
allows multiple callers of `body()` to parse the request. if the body of the message has previously been parsed, it'll use that copy.

lib/request-proxy.js
jwolski: yea, this is a hot path that'll need careful attn. ill fix it in a different PR.

Raynos:
@mranney 

 - The http sharding logic needs to peak into the body.
 - Ringpop needs to parse the body to proxy
 - Sirvice needs to parse the body to do json validation and stuff.

because multiple seperate modules all need the body and its a single stream we are using body caching so that `body(req, res, { cache: true }, cb)` works if called multiple times.

Caveat it will only work if called sequentially, we dont enable parallel caching.

PR #18, opened 2015-02-03
Make ringpopKeys an array

markyen:
* Renames ringpopKey to ringpopKeys
  * Makes ringpopKeys an array of strings instead of a single string
  * Renames checksum to ringpopChecksum

@jwolski @Raynos


jwolski:
lgtm

Raynos:
lgtm.

PR #19, opened 2015-02-04
Handle or proxy all

markyen:
Takes a request with multiple keys and fans it out to all the destinations responsible

@jwolski @Raynos

### DIFF
--- a/index.js
+++ b/index.js
@@ -848,29 +848,32 @@ RingPop.prototype.handleOrProxyAll =
         var keys = opts.keys;
         var req = opts.req;
         var localHandler = opts.localHandler;
+        var whoami = this.whoami();
 
-        var reses = {};
         var dests = mapUniq(keys, this.lookup.bind(this));
-        dests.forEach(function(dest) {
+        return dests.reduce(function(responses, dest) {
             var res = hammock.Response();
-            if (self.whoami() === dest) {
+            if (whoami === dest) {
+                self.logger.trace('handleOrProxyAll was handled', {
+                    keys: keys,
+                    url: req && req.url
+                });
                 process.nextTick(localHandler.bind(null, req, res));
             } else {
-                process.nextTick(self.proxyReq.bind(self, _.defaults({
-                    dest: dest,
-                    res: res
-                }, opts)));
+                self.logger.trace('handleOrProxyAll was proxied', {
+                    keys: keys,
+                    url: req && req.url
+                });
+                process.nextTick(self.proxyReq.bind(self, {
+                    keys: keys,
+                    req: req,
+                    res: res,
+                    dest: dest
+                }));
             }
-            reses[dest] = res;
-        });
-        return reses;
-
-        function mapUniq(list, iteratee) {
-            return Object.keys(list.reduce(function(acc, val) {
-                acc[iteratee(val)] = null;
-                return acc;
-            }, {}));
-        }
+            responses[dest] = res;
+            return responses;
+        }, {});
     };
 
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "view-cover": "opn coverage/index.html"
   },
   "dependencies": {
-    "body": "^4.6.0",
+    "body": "Raynos/body#5308bb4",
```
Raynos: we cannot ship github links into prod. We'll need to wait for a new version of body.

Raynos:
lgtm. Let's wait for body though.

index.js
```diff
@@ -841,6 +842,37 @@ RingPop.prototype.handleOrProxy =
...
+                process.nextTick(self.proxyReq.bind(self, _.defaults({
+                    dest: dest,
+                    res: res
+                }, opts)));
+            }
+            reses[dest] = res;
+        });
+        return reses;
+
+        function mapUniq(list, iteratee) {
```
jwolski: if I were to channel my inner Matt Ranney, can you extract this method to ringpop's prototype to reduce GC?

index.js
```diff
@@ -841,6 +842,37 @@ RingPop.prototype.handleOrProxy =
...
+        var self = this;
+        var keys = opts.keys;
+        var req = opts.req;
+        var localHandler = opts.localHandler;
+
+        var reses = {};
+        var dests = mapUniq(keys, this.lookup.bind(this));
+        dests.forEach(function(dest) {
+            var res = hammock.Response();
+            if (self.whoami() === dest) {
```
jwolski: extract this from the loop and call whoami() only once

index.js
```diff
@@ -841,6 +842,37 @@ RingPop.prototype.handleOrProxy =
...
+                process.nextTick(localHandler.bind(null, req, res));
+            } else {
+                process.nextTick(self.proxyReq.bind(self, _.defaults({
+                    dest: dest,
+                    res: res
+                }, opts)));
```
jwolski: is there a reason not to be explicit with the opts that we're sending to proxyReq and just providing all of `opts`? This'll pass over `localHandler` which is obviously unnecessary.

jwolski:
seems legit. minor comments left

index.js
```diff
@@ -841,6 +842,37 @@ RingPop.prototype.handleOrProxy =
...
+RingPop.prototype.handleOrProxyAll =
+    function handleOrProxyAll(opts) {
+        var self = this;
+        var keys = opts.keys;
+        var req = opts.req;
+        var localHandler = opts.localHandler;
+
+        var reses = {};
+        var dests = mapUniq(keys, this.lookup.bind(this));
+        dests.forEach(function(dest) {
```
jwolski: looks like a reasonable use case for:

```
var reses = dests.reduce(...);
```

ill leave it up to you.

index.js
Raynos: `mapUniq` doesn't need `self` or any closure variables. Just put it in the file as a top level function declaration, no need to be on the prototype.

jwolski: that's true too

index.js
markyen: I like that better too. Fixed

index.js
markyen: Fixed

index.js
markyen: Fixed

index.js
markyen: Fixed

jwolski:
looks great. feel free to self-merge and then plz minor version bump and push tag. thx

PR #20, opened 2015-02-09
Upgrades to body@5.0.0

markyen:
@Raynos @jwolski

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "view-cover": "opn coverage/index.html"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


jwolski:
lgtm. plz patch version bump

Raynos:
lgtm. 

@jwolski technically its a major bump. Anyone relying on the `body@4` caching will be broken, i.e. ringpop-http-shard. Implicit peer dependencies :/

jwolski:
listen to jake

PR #21, opened 2015-02-10
Allow stats hooks to be registered for ringpop extensions

jwolski:
@jmccarthy14 @Raynos This'll allow ringpop serial to return stats over the ringpop /admin/stats endpoint

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,11 +24,11 @@ var EventEmitter = require('events').EventEmitter;
 var fs = require('fs');
 var hammock = require('uber-hammock');
 var metrics = require('metrics');
-var TypedError = require('error/typed');
 
 var AdminJoiner = require('./lib/swim').AdminJoiner;
 var createRingPopTChannel = require('./lib/tchannel.js').createRingPopTChannel;
 var Dissemination = require('./lib/members').Dissemination;
+var errors = require('./lib/errors.js');
 var Gossip = require('./lib/swim.js').Gossip;
 var HashRing = require('./lib/ring');
 var Membership = require('./lib/members').Membership;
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -44,12 +44,31 @@ var HOST_PORT_PATTERN = /^(\d+.\d+.\d+.\d+):\d+$/;
 var MAX_JOIN_DURATION = 300000;
 var PROXY_REQ_PROPS = ['keys', 'dest', 'req', 'res'];
 
+var ArgumentRequiredError = TypedError({
```
Raynos: suggestion: consider placing these errors in an errors.js file

This makes all the errors visible in a single file, might be more friendly.

jwolski: yep, agreed. i feel we're getting to that point where it'd be useful to have such a file.

index.js
```diff
@@ -813,6 +860,26 @@ RingPop.prototype.proxyReq = function proxyReq(opts) {
     this.requestProxy.proxyReq(opts);
 };
 
+RingPop.prototype.registerStatsHook = function registerStatsHook(hook) {
+    if (!hook) {
+        throw ArgumentRequiredError({ argument: 'hook' });
+    }
+
+    if (!hook.name) {
+        throw FieldRequiredError({ argument: 'hok', field: 'name' });
```
Raynos: typo: 'hook'

index.js
```diff
@@ -813,6 +860,26 @@ RingPop.prototype.proxyReq = function proxyReq(opts) {
...
+    }
+
+    if (typeof hook.getStats !== 'function') {
+        throw MethodRequiredError({ argument: 'hook', method: 'getStats' });
+    }
+
+    if (this.statsHooks[hook.name]) {
```
Raynos: consider using the `.isStatsHookRegistered()` method.

test/index_test.js
```diff
@@ -190,3 +190,89 @@ test('no opts does not break handleOrProxy', function t(assert) {
...
+    ringpop.registerStatsHook({
+        name: 'myhook',
+        getStats: function getIt() {
+            return {
+                numQueues: 10
+            };
+        }
+    });
+
+    assert.ok(ringpop.getStats().hooks.myhook, 'has hook stats');
```
Raynos: suggestion: add an assertion to see that `myhook` deep equals { numQueues: 10 }

index.js
jwolski: lovely idea

test/index_test.js
jwolski: legit

PR #22, opened 2015-02-12
Simplifies handleOrProxyAll interface by making it async instead of sync

markyen:
@jwolski @Raynos

### DIFF
--- a/index.js
+++ b/index.js
@@ -843,26 +843,35 @@ RingPop.prototype.handleOrProxy =
     };
 
 RingPop.prototype.handleOrProxyAll =
-    function handleOrProxyAll(opts) {
+    function handleOrProxyAll(opts, cb) {
         var self = this;
         var keys = opts.keys;
         var req = opts.req;
         var localHandler = opts.localHandler;
-        var whoami = this.whoami();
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -871,9 +880,19 @@ RingPop.prototype.handleOrProxyAll =
                     dest: dest
                 }));
             }
-            responses[dest] = res;
-            return responses;
-        }, {});
+        });
+
+        function onResponse(err, data, dest) {
+            if (data) {
```
Raynos: you cant just add a dest & keys field to data. You want `{ body: data, dest, keys }`

markyen: Fair enough

index.js
```diff
@@ -843,37 +843,56 @@ RingPop.prototype.handleOrProxy =
     };
 
 RingPop.prototype.handleOrProxyAll =
-    function handleOrProxyAll(opts) {
+    function handleOrProxyAll(opts, cb) {
```
jwolski: lgtm, but to be comprehensive make sure that `cb()` is called even with 0 keys?

Raynos:
lgtm.

index.js
markyen: Good catch!

PR #23, opened 2015-02-13
Fix protocol period timing bug by updating the protocol timing metric af...

jwolski:
...ter every ping

This was effectively a one-line change that has been broken since the move over to tchannel. This bug caused the protocol period to be around 400ms stable-state instead of 200ms. The reason is because the `timing` metric wasn't updated after ping responses were received. This was subtle logic baked into the HTTP transport. I need to harden this against future issues. But I'm punting on that for now.

I took this opportunity to clean some code out of `index.js` and move it into `Gossip`. I think this separates concerns better. Everything having to do with the gossip protocol period is managed by the Gossip abstraction.

The one-line fix is literally the line that now reads: `this.protocolTiming.update(new Date() - pingStartTime);`

@Raynos @mranney

### DIFF
--- a/index.js
+++ b/index.js
@@ -89,10 +89,6 @@ function RingPop(options) {
     this.pingReqTimeout = 5000;
     this.pingTimeout = 1500;
     this.proxyReqTimeout = options.proxyReqTimeout || 30000;
-    this.minProtocolPeriod = 200;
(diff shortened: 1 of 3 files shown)


Raynos:
lgtm.

PR #24, opened 2015-02-16
Remove localHandler in favor of emitting an event

markyen:
@jwolski @Raynos


Raynos:
lets add a test for the request event.

markyen:
The current test does test this case. There is a listener registered at the top of `allocCluster`, and the fact that is is called is asserted at `handle_or_proxy_test.js:87`

Raynos:
Ok, cool, lgtm.

PR #25, opened 2015-02-17
Reduce membership update log spam

jwolski:


index.js
```diff
@@ -480,90 +489,88 @@ RingPop.prototype.issueMembershipChanges = function issueMembershipChanges(check
...
+};
+
 RingPop.prototype.onMembershipUpdated = function onMembershipUpdated(updates) {
     var self = this;
 
-    var updateHandlers = {
-        alive: function onAliveMember(member) {
-            /* jshint camelcase: false */
-            self.stat('increment', 'membership-update.alive');
-            self.logger.info('member is alive', {
```
Raynos: Your allowed to downgrade this to a `.trace()`.

The latests logtron configures the trace level to log nowhere by default and must be opt into by the application developer.

This allows you to leave very verbose logging in place that could be turned on with a redeploy of the app config.

lib/members.js
```diff
@@ -399,8 +435,99 @@ Membership.prototype._emitUpdated = function _emitUpdated(updates) {
     this.emit('updated', updates);
 };
 
+function MembershipUpdateTracker(options) {
```
jwolski: I'm still writing tests for this guy

test/swim_test.js
```diff
@@ -191,14 +192,14 @@ test('suspicion subprotocol cannot be reenabled without all timers first being s
...
-    assert.plan(2);
+    assert.plan(1);
 
     var suspicion = createSuspicion();
     var member = suspicion.ringpop.membership.remoteMember;
     suspicion.setTimeout = function(fn) { return fn(); };
-    suspicion.ringpop.membership.update = function(changes) {
-        assert.equals(changes[0].address, member.address, 'updates correct member');
-        assert.equals(changes[0].status, 'faulty', 'marks member faulty');
+    // TODO Brittle test, just use real membership object
```
Raynos: Generally when writing tests, i tend to not mock out any logic, even if it's a separate unit.

This does make your tests a hybrid unit/integration test instead of a pure unit test, but most of the times you'd rather test all the logic and avoid having to write the unit twice, once in the code and once in the mock.

I tend to only mock out IO or thin clients around IO. `tchannel` would be a good thing to mock out as it's a thin client around IO.

lib/members.js
```diff
@@ -399,8 +435,99 @@ Membership.prototype._emitUpdated = function _emitUpdated(updates) {
...
+};
+
+MembershipUpdateTracker.prototype.track = function track(updates) {
+    if (!Array.isArray(updates) || updates.length === 0) {
+        return;
+    }
+
+    var sinceLastUpdate = Date.now() - this.lastUpdateTime;
+    if (sinceLastUpdate >= this.flushInterval) {
+        this.flushBuffer();
```
Raynos: should we check `this.anyUpdates()` before calling `flushBuffer()`

Raynos: You already check inside `flushBuffer()` so not needed.

lib/members.js
```diff
@@ -399,8 +435,99 @@ Membership.prototype._emitUpdated = function _emitUpdated(updates) {
...
+        return total + self.buffer[key].length;
+    }, 0);
+};
+
+MembershipUpdateTracker.prototype.renewFlushTimer = function renewFlushTimer() {
+    clearTimeout(this.flushTimer);
+
+    var self = this;
+    this.flushTimer = setTimeout(function onTimeout() {
+        if (self.anyUpdates()) {
```
Raynos: You check for anyUpdates inside `flushBuffer()` so its not needed here.

lib/members.js
Raynos: this class just tracks stuff in memory. I presume it will be used to read from in some future PR.

test/swim_test.js
jwolski: yes, i dont like it either. was a bad decision made by me. i have to remove these kinds of tests when time allows

Raynos: no worries, I recommend pay as you go refactoring when you touch them.

Extra tests don't hurt as long as they don't break, just refactor the ones that break.

PR #27, opened 2015-02-18
Add config hooks to allow alternate source to provide different values a...

jwolski:
@Raynos @iproctor @markyen @sunweik @jmccarthy14 Allow us to easily tune the configurables defined in this PR in production at runtime.


jwolski:
Based on prior conversations, I'm closing this out of core ringpop and moving it into our client layer.

PR #29, opened 2015-02-20
Rewrite join cluster

jwolski:
This is a rewrite of join cluster. It continuously selects from a group of preferred and non-preferred bootstrap hosts until `joinSize` is satisfied or `maxJoinDuration` is exceeded. Joins occur in parallel and allow for an optional `parallelismFactor` which is a multiple of the `joinSize` in case you want more parallelism than `joinSize`.

Still TODO:
1) More test coverage
2) More efficient `init()` function


jwolski:
Since there's still some TODOs, I'm closing until really really ready so as not to confuse people.

PR #30, opened 2015-02-22
Rewrite of join cluster, with parallelism and better node selection.

jwolski:
This is a re-implementation of the join-cluster operation performed on startup of a ringpop instance. This was necessary for a few reasons, but mostly because I noticed that bootstrap times in production were higher than desired, anywhere between 200ms and 1s. This is unacceptably long. Improvements made in this PR will improve that and allow us to iterate on other solutions faster.

The original implementation would join 3 other instances (from the bootstrap list) in serial and would randomly choose nodes from that list in a naive way. It would/could choose the same instance over and over again even if a join attempt had previously failed.

This re-implementation makes the following improvements:
1) makes the join parallel
2) offers a join parallelism factor that allows joins to be performed across more nodes than required to fulfill `joinSize` in attempt to make the join complete faster.
3) makes better choices for whom to join: preferred nodes vs. non-preferred. Preferred nodes are those that are not on the same host as the current instance which is suitable for deploys and host-wide restarts. While this likely won't suit all use cases, it'll be a good first effort. Maybe at some point which can make pluggable "preferred node"-strategies, but at the moment, it's unnecessary.

@Raynos @robskillington @mranney @jmccarthy14 @jcorbin


lib/join_cluster.js
```diff
@@ -0,0 +1,396 @@
...
+    type: 'ringpop.join-duration-exceeded',
+    message: 'Join duration of `{duration}` exceeded max `{max}`',
+    duration: null,
+    max: null
+});
+
+var JOIN_SIZE = 3;
+var JOIN_TIMEOUT = 1000;
+var MAX_JOIN_DURATION = 120000;
+var PARALLELISM_FACTOR = 2;
```
robskillington: Out of curiousity what is likely a good lower/max bound on how much parallelism to use for joining?  Would be interesting to be able to run some benchmarks and see how different values seem to perform.

Raynos:
Is this ready for review? If so, can we start a habit of ccing reviewers in the description field of the pull request.

A PR with no description or cc'ing of reviewers might just be a pre-emptive not-ready for review PR.

jwolski:
yep. this is generally ready for review. the habit of ccing reviewers is already well established. i made this public over the weekend in prep for eventual ccing of relevant people. even though the code has its own comments, youre right in pointing out that this pr deserves its own detailed description and inclusion into the readme. thatll be taken care of tomorrow.

lib/join_cluster.js
jwolski: Very good question. This impl gives us some room to play. I eventually want to get us to a point in production where these values are dynamic and tunable at runtime.

index.js
```diff
@@ -86,7 +86,7 @@ function RingPop(options) {
     this.isReady = false;
 
     this.debugFlags = {};
-    this.joinSize = 3;              // join fanout
+    this.joinSize = options.joinSize;
```
Raynos: consider defaulting this to something or making it a required field. just like app & hostPort are required.

index.js
```diff
@@ -180,13 +180,11 @@ RingPop.prototype.adminJoin = function adminJoin(target, callback) {
         this.joiner = null;
     }
 
-    this.joiner = new AdminJoiner({
+    this.joiner = joinCluster({
```
Raynos: nice. Liking the `fn(options, cb)` pattern.

index.js
```diff
@@ -278,7 +276,7 @@ RingPop.prototype.bootstrap = function bootstrap(bootstrapFile, callback) {
         self.isReady = true;
         self.emit('ready');
 
-        if (callback) callback();
+        if (callback) callback(err, nodesJoined);
```
Raynos: nitpick: doing `callback(null, nodesJoined)` makes it slightly clearer then we early returned for all the error cases

index.js
```diff
@@ -889,6 +892,14 @@ RingPop.prototype.handleOrProxyAll =
         }
     };
 
+RingPop.prototype.allowJoins = function allowJoins() {
```
Raynos: Are the `allowJoins` and `denyJoins` just helpers for testing ? We should consider leaving a comment about that.

Or is there an application / production use case for these methods ?

test/integration/admin_join_test.js
```diff
@@ -24,15 +24,13 @@ var test = require('tape');
 
 var allocRingpop = require('../lib/alloc-ringpop.js');
 
-test('bootstrap with self', function t(assert) {
+test('bootstrap with self is ok', function t(assert) {
```
Raynos: nice! This is a useful feature :)

test/integration/join_cluster_test.js
```diff
@@ -0,0 +1,215 @@
...
+'use strict';
+
+var _ = require('underscore');
+var after = require('after');
+var DebuglogLogger = require('../../lib/debug-log-logger.js');
+var Ringpop = require('../../index.js');
+var TChannel = require('tchannel');
+var test = require('tape');
+
+function bootstrapClusterOf(size, opts, tap, onBootstrap) {
```
Raynos: suggestion: move this function and related guys into something like https://github.com/uber/ringpop/blob/master/test/lib/alloc-cluster.js or create a new helper to set up a cluster of ringpops. 

Seems like something you want to write once and re-use for all integration tests.

test/integration/join_cluster_test.js
```diff
@@ -0,0 +1,215 @@
...
+
+function createRingpop(opts) {
+    opts = opts || {};
+
+    var ringpop = new Ringpop(_.extend({
+        app: 'test',
+        hostPort: opts.host + ':' + opts.port,
+        maxJoinDuration: opts.maxJoinDuration,
+        channel: createTChannel(opts.host, opts.port),
+        logger: DebuglogLogger()
```
Raynos: suggestion: There is a module called `DebugLogtron` (`require('debug-logtron')`) that is basically the same as `DebuglogLogger` and can remove a bit of code weight from ringpop.

test/integration/join_cluster_test.js
```diff
@@ -0,0 +1,215 @@
...
+    return ringpop;
+}
+
+function destroyCluster(cluster) {
+    cluster.forEach(function eachRingpop(ringpop) {
+        ringpop.destroy();
+    });
+}
+
+test('one node can join', function t(assert) {
```
Raynos: suggestion: consider having a `bootstrapCluster.test` helper.

```js
bootstrapCluster.test = function test(name, opts, fn) {
    test(name, function onAssert(assert) {
        assert.once('end', onEnd);

        var cluster = bootstrapClusterOf(
            opts.size, opts, opts.tap, onBootstrap
        );

        function onEnd() {
            destroyCluster(cluster);
        }

        function onBootstrap(err) {
            if (err) {
                assert.ifError(err);
                return assert.end();
            }

            fn(cluster, assert);
        }
    });
};


bootstrapCluster.test('one node can join', {
    size: 1
}, function t(cluster, assert) {
    assert.end();
});
```

Thank @jcorbin for this pattern.

test/integration/join_cluster_test.js
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    var cluster = bootstrapClusterOf(3, {
+        joinSize: 1
+    }, function tap(cluster) {
+        cluster[2].denyJoins();
+    }, onBootstrap);
+});
+
+test('three nodes, 1 bad node, join size 2', function t(assert) {
```
Raynos: note: this has two bad nodes in it.

test/integration/join_cluster_test.js
```diff
@@ -0,0 +1,215 @@
...
+        destroyCluster(cluster);
+        assert.end();
+    });
+
+    function onBootstrap(err) {
+        assert.ifError(err, 'no bootstrap error occurred');
+        done();
+    }
+
+    var cluster = bootstrapClusterOf(25, null, null, onBootstrap);
```
Raynos: do we want to add an assert for join convergence time ?

test/integration/join_cluster_test.js
```diff
@@ -0,0 +1,215 @@
...
+
+    function onSlowBootstrap(err) {
+        assert.equal(err.type, 'ringpop.join-aborted', 'join aborted error');
+        done();
+    }
+
+    bootstrapClusterOf(2, null, function tap(cluster) {
+        cluster[1].protocolJoin = function protocolJoin(options, callback) {
+            setTimeout(function onTimeout() {
+                destroyCluster(cluster);
```
Raynos: destroyign here feels premature, maybe destroy it just before `assert.end()` ?

lib/join_cluster.js
```diff
@@ -0,0 +1,402 @@
...
+var JoinDurationExceededError = TypedError({
+    type: 'ringpop.join-duration-exceeded',
+    message: 'Join duration of `{duration}` exceeded max `{max}`',
+    duration: null,
+    max: null
+});
+
+var JOIN_SIZE = 3;
+var JOIN_TIMEOUT = 1000;
+var MAX_JOIN_DURATION = 120000;
```
Raynos: 2 minutes feels like a long time. Maybe time out after 30 seconds instead.

lib/join_cluster.js
```diff
@@ -0,0 +1,402 @@
...
+            return callback(err);
+        }
+
+        nodesJoined = nodesJoined.concat(nodes.successes);
+        numJoined += nodes.successes.length;
+        numFailed += nodes.failures.length;
+        numGroups++;
+
+        if (numJoined >= self.joinSize) {
+            self.ringpop.logger.info('ringpop join complete', {
```
Raynos: suggest incrementing statsd counters

Raynos:
lgtm. Mostly suggestions on how to cleanup the tests.

index.js
jwolski: it's defaulted in `join_cluster.js` which is a bit odd, but since `index.js` is the top-level ringpop object it's a temporary holding place for these values.

index.js
jwolski: agreed

index.js
jwolski: just for testing. agreed re: commenting.

lib/join_cluster.js
jwolski: 2 minutes is a bit long. if a join can't take place within a few seconds, there's usually something the matter. the reason for the longer time period is because it may give people a bit more time to look to see what's wrong without having to do a full restart of all instances. either way, 30s or 120s is rather arbitrary.

lib/join_cluster.js
jwolski: hmmmm, sure.

test/integration/join_cluster_test.js
jwolski: will do

test/integration/join_cluster_test.js
jwolski: will do

test/integration/join_cluster_test.js
jwolski: nice. ill play with it.,

test/integration/join_cluster_test.js
jwolski: thanks

test/integration/join_cluster_test.js
jwolski: probably not in this test. i think that's a separate concern.

test/integration/join_cluster_test.js
jwolski: destroying here causes the abort. it's part of the test. ill put a comment. since this is non-obvious, ill put a comment.

lib/join_cluster.js
Raynos: Interesting having a fail in 5s; shutdown in 120s is an interesting pattern.

This gives a 2 minute window of in production debugging.

Leave a comment about that :)

PR #31, opened 2015-02-23
Request proxy retries

jwolski:
Proxied requests suffer from cluster inconsistencies on deploys and what not. Build retries w/ backoff into the request proxying layer to deal with an eventually convergent ring. @Raynos recommended I use a module for backoff, but I decided against it due to the simplicity of the approach below. Feel free to feedback.

@Raynos @jmccarthy14 @mranney @jsu1212 @jcorbin

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,31 +59,23 @@ ringpop.on('changed', function() {
 });
 ```
 
-# Forwarding a request
-Ringpop will typically be used by handing over request routing to it. As described earlier, the "process or forward" pattern is used to decide whether a request should be processed by the node that received the request or by another node. As an alternative, `handleOrForward` can be used to encapsulate that repetitive pattern. Here's an example of its use:
+# Request Proxying
+ringpop provides request routing as a convenience. When a request arrives at your services' public endpoints, you'll want to use ringpop to decide whether request processing should take place on the node that received the request or elsewhere. If elsewhere, ringpop will proxy your request to the correct destination.
 
-```javascript
-// Let's say this is an endpoint handler in your web application
-function endpoint(incoming, opts, cb) {
-    function handle() {
-        // process request
-    }
+Upon arrival of a proxied request at its destination, membership checksums of the sender and receiver will be compared. The request will be refused if checksums differ. Mismatches are to be expected when nodes are entering or exiting the cluster due to deploys, added/removed capacity or failures. The cluster will eventually converge on one membership checksum, therefore, refused requests are best handled by retrying them.
 
-    function forwarded(err, resp, body) {
-        cb(err, {
-            statusCode: resp && resp.statusCode
-        });
-    }
+ringpop's request proxy has retries built in and can be tuned using 2 parameters provided at the time ringpop is instantiated: `requestProxyMaxRetries` and `requestProxyRetrySchedule` or per-request with: `maxRetries` and `retrySchedule`. The first parameter is an integer representing the number of times a particular request is retried. The second parameter is an array of integer or floating point values representing the delay in between consecutive retries.
 
-    var requestToForward = {
-        method: 'POST',
-        path: '/supply/1',
-        headers: { 'Content-Type: application/json' },
-        body: JSON.stringify({ /* fill body here */ }),
-        timeout: 1000
-    };
+ringpop has codified the aforementioned routing pattern in the `handleOrProxy` function. It returns true when `key` hashes to the "current" node and false otherwise. `false` results in the request being proxied to the correct destination. Its usage looks like this:
 
-    ringpop.handleOrForward(incoming.params.uuid, handle, requestToForward, forwarded);
+```js
+var opts = {
+    maxRetries: 3,
+    retrySchedule: [0, 0.5, 2]
+};
+
+if (ringpop.handleOrProxy(key, req, res, opts)) {
+    // handle request
 }
 ```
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -59,31 +59,23 @@ ringpop.on('changed', function() {
...
+Upon arrival of a proxied request at its destination, membership checksums of the sender and receiver will be compared. The request will be refused if checksums do not match. Mismatches are to be expected when nodes are entering or exiting the cluster due to deploys, added/removed capacity or failures. The cluster will eventually converge on one membership checksum, therefore, refused requests are best handled by retrying them.
+
+ringpop's request proxy has retries built in and can be tuned using 2 parameters provided at the time ringpop is instantiated: `requestProxyMaxRetries` and `requestProxyRetrySchedule` or per-request with: `maxRetries` and `retrySchedule`. The first parameter is an integer representing the number of times a particular request is retried. The second parameter is an array of integer or floating point values representing the delay in between consecutive retries.
+
+ringpop has codified the aforementioned routing pattern in the `handleOrProxy` function. It returns true when `key` hashes to the "current" node and false otherwise. `false` results in the request being proxied to the correct destination. Its usage looks like this:
+
+```js
+var opts = {
```
Raynos: can we add statsd for retries. This allows us to see on a dashboard how much retrying ringpop does internally for proxying.

In fact adding statsd full stop so we can see proxy health in statsd would be super useful.

jwolski: yes absolutely. i though the same after creating the pr. will add it in tomorrow.

lib/request-proxy.js
```diff
@@ -26,6 +26,8 @@ var TypedError = require('error/typed');
 
 var safeParse = require('./util').safeParse;
 
+var RETRY_SCHEDULE = [0, 0.5, 2];
```
Raynos: will ringpop converge in 500ms on a bad checksum ?

lib/request-proxy.js
```diff
@@ -122,6 +149,77 @@ proto.proxyReq = function proxyReq(opts) {
...
+        // Per-request parameters override instance-level
+        maxRetries = numOrDefault(maxRetries, self.maxRetries);
+        retrySchedule = retrySchedule || self.retrySchedule;
+
+        var numRetries = 0;
+        var reqStartTime = Date.now();
+        var retryStartTime = null;
+
+        return function sendIt(options, head, body, callback) {
+            if (self.maxRetries === 0) {
```
Raynos: use `maxRetries` instead of `self.maxRetries`.

lib/request-proxy.js
```diff
@@ -122,6 +149,77 @@ proto.proxyReq = function proxyReq(opts) {
...
+                ringpop.channel.send(options, '/proxy/req',
+                    head, body, callback);
+            } else {
+                ringpop.channel.send(options, '/proxy/req', head, body,
+                    function onSend(err, res1, res2) {
```
Raynos: suggestion: lets move this closure up a level by making it a function declaration (less indents!)

lib/request-proxy.js
```diff
@@ -122,6 +149,77 @@ proto.proxyReq = function proxyReq(opts) {
...
+
+        var numRetries = 0;
+        var reqStartTime = Date.now();
+        var retryStartTime = null;
+
+        return function sendIt(options, head, body, callback) {
+            if (self.maxRetries === 0) {
+                ringpop.channel.send(options, '/proxy/req',
+                    head, body, callback);
+            } else {
```
Raynos: suggestion: can reduce indentation by early returning in `if` case and not using `else`.

lib/request-proxy.js
Raynos: Is doing a retry with 0 delay bad? It's probably not good to do an immediate retry on failure. Not sure what a good delay is though.

jwolski: it very well could, but maybe initially we be a little bit more conversative and bump up the max to 2+

lib/request-proxy.js
jwolski: good catch

lib/request-proxy.js
jwolski: k

lib/request-proxy.js
jwolski: good idea

lib/request-proxy.js
```diff
@@ -36,19 +38,43 @@ var InvalidCheckSumError = TypedError({
...
+
+function RequestProxy(opts) {
+    this.ringpop = opts.ringpop;
+    this.retrySchedule = opts.retrySchedule || RETRY_SCHEDULE;
+    this.maxRetries = numOrDefault(opts.maxRetries, this.retrySchedule.length);
+
+    // Simply a convenience over having to index into `retrySchedule` each time
+    this.maxRetryTimeout = this.retrySchedule[this.retrySchedule.length - 1];
+
+    this.retryTimer = null;
```
Raynos: this field cannot be a single value. Multiple `proxyReq()` methods will assign to it at the same time. 

This needs to be an object or array of pending timers. Destroy will have to loop over them.

jwolski: good catch

Raynos:
Finished reviewing. Noted a couple of points. 

I really like the logging in this one :)

Raynos:
Looking good. Final note: lets update the README with the new stats & events.

jwolski:
yep, already done. just havent pushed it up. ive also gone down the
rabbithole or re-routing requests if lookup() has changed hosts :|

Raynos:
lgtm. #rebase and #shipit

PR #32, opened 2015-02-24
Split changed event into two: membershipChanged and ringChanged

jwolski:
This change is necessary for the sevnups of the world that are looking to take action when ring-state changes (node added/removed). The problem today is that ringpop will fire events for both membership changes and ring changes under the same name causing sevnup to rescan the vnode-space for membership updates like when node are suspected. This is not a huge problem, but splitting events improves correctness.

@iproctor @Raynos @jmccarthy14

### DIFF
--- a/index.js
+++ b/index.js
@@ -400,10 +400,7 @@ RingPop.prototype.protocolJoin = function protocolJoin(options, callback) {
     this.serverRate.mark();
     this.totalRate.mark();
 
-    this.membership.addMember({
-        address: joinerAddress,
-        incarnationNumber: options.incarnationNumber
-    });
+    this.membership.makeJoin(joinerAddress, options.incarnationNumber);
 
     callback(null, {
         app: this.app,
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -400,10 +400,7 @@ RingPop.prototype.protocolJoin = function protocolJoin(options, callback) {
     this.serverRate.mark();
     this.totalRate.mark();
 
-    this.membership.addMember({
-        address: joinerAddress,
-        incarnationNumber: options.incarnationNumber
-    });
+    this.membership.makeJoin(joinerAddress, options.incarnationNumber);
```
jwolski: Incidental refactor to be more consistent with `makeFaulty`, `makeSuspect`, etc operations

index.js
```diff
@@ -580,23 +577,35 @@ RingPop.prototype.onMemberSuspect = function onMemberSuspect(member) {
 
 RingPop.prototype.onMembershipUpdated = function onMembershipUpdated(updates) {
     var self = this;
+    var membershipChanged = false;
+    var ringChanged = false;
 
     updates.forEach(function(update) {
         if (update.type === 'alive') {
             self.onMemberAlive(update);
+            ringChanged |= membershipChanged |= true;
```
Raynos: `|=` is for doing bit operations on numbers. It's weird to use the "coercion" feature of javascript for this.

We should use the long form. i.e. `ringChanged = true; membershipChanged = true;`

Raynos:
lgtm. other then `|=` is kind of weird.

index.js
jwolski: true. it's a habit and since there's no one doing `|= false` the bit-wise operators are unnecessary. i'll take it out.

PR #33, opened 2015-02-26
Request proxy endpoint override

jwolski:
@Raynos @markyen This will allow proxy requests to override the proxy endpoint.

### DIFF
--- a/index.js
+++ b/index.js
@@ -881,12 +881,12 @@ RingPop.prototype.handleOrProxyAll =
                     url: req && req.url,
                     dest: dest
                 });
-                self.proxyReq({
+                self.proxyReq(_.defaults({
                     keys: keys,
                     req: req,
                     res: res,
                     dest: dest
-                });
+                }, opts));
             }
         });
 
--- a/test/integration/proxy_req_test.js
+++ b/test/integration/proxy_req_test.js
@@ -23,6 +23,7 @@ var after = require('after');
 var test = require('tape');
 
 var allocCluster = require('../lib/alloc-cluster.js');
+var serializeHead = require('../../lib/request-proxy-util.js').serializeHead;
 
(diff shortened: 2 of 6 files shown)


test/integration/proxy_req_test.js
```diff
@@ -287,3 +288,73 @@ test('removes some timeouts, not others', function t(assert) {
...
+            json: { hello: true },
+            maxRetries: 0
+        }, function onRequest(err, resp) {
+            assert.deepEqual(resp.body, things, 'responds with body');
+
+            cluster.destroy();
+            assert.end();
+        });
+
+        var head = serializeHead(request, {
```
Raynos: nitpick. it reads easier if we move this above cluster.two.channel.register()

Raynos:
lgtm.

test/integration/proxy_req_test.js
jwolski: can't do that because of `request` used as first param. i originally assigned `head` within the `register` call, but that seemed a little too late.

Raynos: oh i get it now. cool makes sense.

index.js
```diff
@@ -881,12 +881,12 @@ RingPop.prototype.handleOrProxyAll =
                     url: req && req.url,
                     dest: dest
                 });
-                self.proxyReq({
+                self.proxyReq(_.defaults({
                     keys: keys,
                     req: req,
```
markyen: `keys` and `req` are already in `opts`, so you could leave them out here if you don't want to set them twice

jwolski: yea, this was kind of the lazy approach. I dont want to introspect `opts` internals to the point that I would leave something out and plucking the ones out that you know are already in `opts` will be taken care of by `_.defaults()`

markyen:
lgtm

PR #34, opened 2015-02-26
Stringify our bodies and responses for tchannel.

Raynos:
Tchannel v1 expects to get an object, string or buffer.

Tchannel v2 expects a string or buffer.

If you pass an object to tchannel v1 it will stringify it
for you. We should stringify it explicitely in ringpop so
that ringpop works with either v1 or v2.

reviewers: @jwolski @jcorbin

cc: @jsu1212

### DIFF
--- a/lib/swim.js
+++ b/lib/swim.js
@@ -85,11 +85,11 @@ AdminJoiner.prototype.sendJoin = function sendJoin() {
         timeout: this.ringpop.pingReqTimeout
     };
     var local = this.ringpop.membership.localMember;
-    var body = {
+    var body = JSON.stringify({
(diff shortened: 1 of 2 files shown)


jcorbin:
What of parsing then on the other side? Doesn't ringpop need to do that now too? Or was this always an asymmetric convenience?

jwolski:
lgtm. can you squash and merge plz?

Raynos:
@jcorbin this was always an asymmetric convenience.

Raynos:
ill wait for travis.

Raynos:
screw travis.

jwolski:
lulz

PR #35, opened 2015-02-27
fix: don't ignore --hosts command line argument value

bnugmanov:
The --hosts argument is required, but its value is then ignored. It probably works out most of the time, since the default value of ./hosts.json is there. Still, a bug.

Tested with a seed hosts file value of "hosts1.json". Before fix, main barfs looking for "./hosts.json". 

Works as expected after fix.

Unit tests pass (as per git hooks)

### DIFF
--- a/main.js
+++ b/main.js
@@ -86,7 +86,7 @@ function main() {
     });
 
     createRingPopTChannel(ringpop, tchannel);
-    ringpop.bootstrap();
+    ringpop.bootstrap(program.hosts);
 }
 
 if (require.main === module) {


Raynos:
lgtm.

bnugmanov:
@jwolski @Raynos could you please have a look? Thanks!

jwolski:
yep, lgtm. thanks for the contribution. `main.js` doesn't see a lot of use in production since ringpop is not run standalone. But when running with `tick-cluster.js` a `hosts.json` is automatically generated with each run.

PR #37, opened 2015-03-03
Introduce lower bound on incarnation number for local member asserting its aliveness

jwolski:
Without having a lower bound on the incarnation number, it causes cluster to disseminate more changes than necessary. During a deploy, nodes that detect suspects are disseminating that info throughout the cluster eventually arriving at the original suspect once restarted. When applying the change at that member, incarnation number is bumped a bunch of times until the ring converges. This is apparent in the membership update rollups recently introduced:

```
fields.updates_.10.30.11.2:20801_	  	{"type":"suspect","address":"10.30.11.2:20801","status":"suspect","incarnationNumber":1424847893496,"isLocal":false}, {"type":"alive","address":"10.30.11.2:20801","status":"alive","incarnationNumber":1425346015439,"isLocal":false}, {"type":"alive","address":"10.30.11.2:20801","status":"alive","incarnationNumber":1425346015737,"isLocal":false}, {"type":"alive","address":"10.30.11.2:20801","status":"alive","incarnationNumber":1425346016474,"isLocal":false}
```

Notice:
statuses: suspect -> alive -> alive -> alive -> alive

incarnation numbers: 
1424847893496
1425346015439
1425346015737
1425346016474

@iproctor @Raynos @mrhooray

### DIFF
--- a/lib/member.js
+++ b/lib/member.js
@@ -17,4 +17,17 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-require('glob').sync(__dirname + '/**/*{_test,-test}.js').forEach(require);
+'use strict';
+
+function Member() {
+
+}
+
+Member.Status = {
+    Alive: 'alive',
+    Faulty: 'faulty',
+    Leave: 'leave',
+    Suspect: 'suspect'
+};
+
+module.exports = Member;
--- a/test/lib/test-ringpop.js
+++ b/test/lib/test-ringpop.js
@@ -17,4 +17,35 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-require('glob').sync(__dirname + '/**/*{_test,-test}.js').forEach(require);
+'use strict';
+
+var Ringpop = require('../../index.js');
+var tape = require('tape');
+
+function testRingpop(opts, name, test) {
+    if (typeof opts === 'string' && typeof name === 'function') {
+        test = name;
+        name = opts;
+        opts = {};
+    }
+
+    tape(name, function onTest(assert) {
+        var ringpop = new Ringpop({
+            app: opts.app || 'test',
+            hostPort: opts.hostPort || '127.0.0.1:3000'
+        });
+
+        ringpop.addLocalMember();
+
+        test({
+            localMember: ringpop.membership.localMember,
+            membership: ringpop.membership,
+            ringpop: ringpop
+        }, assert);
+
+        assert.end();
+        ringpop.destroy();
+    });
+}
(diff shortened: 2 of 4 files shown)


mrhooray:
lgtm.

However time/epoch based incarnation number may not work well with time drift caused by ntpd misconfiguration, etc. What about using logical clock as incarnation number? Since we don't care about ordering of events from different nodes and incarnation number can only be bumped by local member, an integer with some modification on handling update to local member may work.

robskillington:
+1 for avoiding the system time.  Seen a lot of bugs caused by drift.

jwolski:
yep. lamport timestamps are an eventual option. but i see drift as a separate issue from misconfigurarion. typically i dont see backwards drift. so long as we have monotonically increasing numbers we should be good. misconfig is another thing.

Raynos:
@jwolski let's have a seperate issue for monotonic Incarnation numbers. Date.now() is not monotonic.

Consider "monotonic-timestamp" or similar approach.

mrhooray:
I should have separated the cases that may cause system time to go backwards like misconfiguration from time drift given the nature of time and our use on incarnation number. Not relying on system time would prevent us from getting hurt from these situations but it may be fine assumption that system time does not go backwards.

robskillington:
"Anything that can go wrong will go wrong" I've seen hosts that flap their date because something like puppet/another configuration system was in place that was constantly trying to self heal by changing timezones and flapping between them =[

jwolski:
yep, agreed that system clock is asking for trouble. this is an issue until we get monotonically increasing clocks that are not susceptible to drift or misconfigs. As @Raynos said, I'll separate this PR from that effort.

lib/member.js
```diff
@@ -0,0 +1,33 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+'use strict';
+
+function Member() {
```
Raynos: consider `var Member = {};`

This means people will not accidentally call `Member()`

test/lib/test-ringpop.js
```diff
@@ -0,0 +1,51 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+'use strict';
+
+var Ringpop = require('../../index.js');
+var tape = require('tape');
+
+function testRingpop(opts, name, test) {
```
Raynos: nice :0

Raynos:
lgtm.

PR #39, opened 2015-03-03
Add timestamps to membership update rollups

jwolski:
@Raynos @mrhooray


Raynos:
lgtm.

mrhooray:
lgtm. xtend is shallow but update seems only contain primitive type value

PR #41, opened 2015-03-03
Lookup latest destination when retrying a proxied request

jwolski:
This is an enhancement to proxied retries. The enhancement addresses the lack of key lookups after proxied requests fail. I've taken this opportunity to abstract away a "send" operation into its own class which encapsulates all of the retry logic. Before a retry is attempted, a lookup is performed on the key. If the destination is different, the request is rerouted to the new destination. Otherwise, not rerouted, but still retried. If multiple keys are part of the send operation (when `handleOrProxyAll` is used) and the key destinations diverge after a request fails, then I abort the retry entirely. Handling this nuanced case will be left as an exercise for later. I've essentially ripped out a lot of what was in `request-proxy.js` and moved it into `request-proxy-send.js`. 

@Raynos @robskillington @mrhooray


test/swim_test.js
```diff
@@ -22,25 +22,13 @@ var mock = require('./mock');
 var test = require('tape');
```
jwolski: These changes are all incidental. Feel free to ignore them. They have nothing to do with `request-proxy-send` introduction.

test/request_proxy_test.js
```diff
@@ -44,9 +44,12 @@ test('request proxy sends custom ringpop metadata in head', function t(assert) {
     var dest = 'disneyworld';
 
     var proxy = createRequestProxy();
-    proxy.ringpop.channel.send = function(options, arg1, arg2, arg3, callback) {
+    var ringpop = proxy.ringpop;
```
jwolski: Incidental changes. Ringpop needs to be destroyed.

lib/request-proxy-send.js
```diff
@@ -0,0 +1,279 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+'use strict';
+
+var numOrDefault = require('./util.js').numOrDefault;
+var rawHead = require('./request-proxy-util.js').rawHead;
```
Raynos: The fact that you have a request-proxy-util file probably means you should have a request-proxy folder.

 - request-proxy/index.js
 - request-proxy/send.js
 - request-proxy/util.js

lib/request-proxy.js
```diff
@@ -86,158 +76,77 @@ proto.proxyReq = function proxyReq(opts) {
...
+    // Request-level overrides (see ctor).
     var maxRetries = opts.maxRetries;
     var retrySchedule = opts.retrySchedule;
     var endpoint = opts.endpoint || '/proxy/req';
-
-    var ringpop = this.ringpop;
-
     var timeout = opts.timeout ?
-        opts.timeout : ringpop.proxyReqTimeout;
+        opts.timeout : this.ringpop.proxyReqTimeout;
```
Raynos: nitpick: consider using `self` and `this` consistently.

You use `this` in the function and `self` in the `onBody` closure.

test/integration/proxy_req_test.js
```diff
@@ -358,3 +358,122 @@ test('overrides /proxy/req endpoint and fails', function t(assert) {
...
+                    return ++count === 1 ? cluster.one.hostPort :
+                        cluster.three.hostPort;
+                };
+            }
+        });
+
+        cluster.requestAll({
+            keys: [cluster.keys.two, cluster.keys.two],
+            host: 'one',
+            maxRetries: 10,
```
Raynos: is it possible to have an `assert.equal(retryCount, 0)` ? 

We want to test early abort but it doesnt look like we have an assert for it.

lib/request-proxy-send.js
```diff
@@ -0,0 +1,279 @@
...
+        ' now meant for {newDestinations}.',
+    keys: null,
+    origDestination: null,
+    newDestinations: null
+});
+
+var RETRY_SCHEDULE = [0, 1, 3.5];
+
+function RequestProxySend(opts) {
+    this.ringpop = opts.ringpop;
```
Raynos: nitpick: I think this pattern of passing the entire `ringpop` interface into the Send operation is not a good pattern.

It would be great if we could pass in just the interface that `RequestProxySend` actually needs.

`{ stat, logger, ring: { whoami, lookup }, emit, channel }`

This way we can see what it actually needs to operate. A `channel` to send() to and a ringpop interface for whoami() and lookup()

lib/request-proxy-send.js
```diff
@@ -0,0 +1,279 @@
...
+    this.errors = [];
+    this.numRetries = 0;
+    this.reqStartTime = Date.now();
+    this.retryStartTime = null;
+    this.timeout = null;
+}
+
+RequestProxySend.prototype.abortRetry = function abortRetry(dests, callback) {
+    this.ringpop.stat('increment', 'requestProxy.retry.aborted');
+    this.ringpop.logger.warn('ringpop request proxy retry aborted after keys destinations diverged', {
```
Raynos: this log statement assumes implementation details that do not match the `abortRetry` method name.

Consider calling this `abortOnKeyDivergence` or changing the log statement

lib/request-proxy-send.js
```diff
@@ -0,0 +1,279 @@
...
+RequestProxySend.prototype.getStrHead = function getStrHead() {
+    return strHead(this.request.obj, {
+        checksum: this.ringpop.membership.checksum,
+        keys: this.keys
+    });
+};
+
+RequestProxySend.prototype.handleMaxRetriesExceeded = function handleMaxRetriesExceeded(callback) {
+    this.ringpop.stat('increment', 'requestProxy.retry.failed');
+    this.ringpop.logger.warn('ringpop request proxy retry exceeded max retries and failed', {
```
Raynos: A lot of these logger statements have a large object.

Maybe in the constructor we should pre-allocate a well defined `logContext` object with the majority of these properties.

This will make the logging context very clear in the constructor and make the actual log sites simpler.

It also allows us to inline a couple of these helper methods.

lib/request-proxy-send.js
```diff
@@ -0,0 +1,279 @@
...
+            destinations: this.destinations
+        });
+
+        this.ringpop.emit('requestProxy.retrySucceeded');
+    }
+
+    callback(null, res1, res2);
+};
+
+RequestProxySend.prototype.lookupKeys = function lookupKeys(keys) {
```
Raynos: nitpick: doing a reduce and Object.keys() feels a bit strange.

You probably wanted

```js
var dests = [];

for (var i = 0; i < keys.length; i++) {
    var key = self.ringpop.lookup(keys[i]);
    if (dests.indexOf(key) === -1) {
        dests.push(key);
    }
}

return dests;
```

lib/request-proxy-send.js
```diff
@@ -0,0 +1,279 @@
...
+        this.retryStartTime = Date.now();
+    }
+
+    // `scheduleDelay` could result in NaN. If so, default to max.
+    var scheduleDelay = this.retries.schedule[this.numRetries] * 1000;
+    var delay = numOrDefault(scheduleDelay, this.maxRetryTimeout);
+
+    // TODO Can we make this delay responsive/adaptive to previously
+    // recorded cluster converge-times?
+    this.timeout = setTimeout(function onTimeout() {
```
Raynos: suggestion: To ease testing you might want to do `self.setTimeout()`

You can then use `time-mock` ( https://github.com/uber/tchannel/blob/master/node/test/timeouts.js#L23 ) to do integration tests without having actual timeouts. This will speed up your integration tests.

Raynos:
Left feedback. However the actual implementation lgtm.

lib/request-proxy-send.js
jwolski: agreed.

lib/request-proxy.js
jwolski: hm. this is subject to personal preference. i usually like using `this` until i know that scope has changed. admittedly, this is error prone with longish functions. ill consider

test/integration/proxy_req_test.js
jwolski: hm, good point. ill try something like it out

lib/request-proxy-send.js
jwolski: This is interesting. I'm not sure if I agree with your suggestion, but I'll let it settle and digest and reevaluate.

lib/request-proxy-send.js
jwolski: good point. thanks

lib/request-proxy-send.js
jwolski: Yea. The verbosity of logging is sometimes invaluable, but it leaves a mess in the codebase. I'll try to come up with some logging context as you suggest.

lib/request-proxy-send.js
jwolski: yep. thanks

lib/request-proxy-send.js
jwolski: thanks

lib/request-proxy.js
Raynos: This is a stylistic thing; agreed. hence a nitpick.

test/integration/proxy_req_test.js
jwolski: addressed

lib/request-proxy.js
jwolski: not being addressed

PR #43, opened 2015-03-05
Do not assume update type for first time member

jwolski:
This is a nasty bug in ringpop. Ringpop will assume that the update type is `alive` no matter what the status of the member is at the time they're added into the membership list.

@Raynos @iproctor @sunweik @robskillington @leizha

### DIFF
--- a/index.js
+++ b/index.js
@@ -161,7 +161,8 @@ RingPop.prototype.setupChannel = function setupChannel() {
 RingPop.prototype.addLocalMember = function addLocalMember(info) {
     this.membership.addMember({
         address: this.hostPort,
-        incarnationNumber: info && info.incarnationNumber
+        incarnationNumber: info && info.incarnationNumber,
+        status: 'alive'
     });
 };
 
@@ -531,23 +532,6 @@ RingPop.prototype.onMemberFaulty = function onMemberFaulty(member) {
     this.suspicion.stop(member);
 };
 
-RingPop.prototype.onMemberJoined = function onMemberJoined(member) {
-    this.stat('increment', 'membership-update.new');
-    this.logger.debug('member has joined', {
-        local: this.membership.localMember.address,
-        joined: member.address
-    });
-
-    this.ring.addServer(member.address);
-
-    this.dissemination.addChange({
-        address: member.address,
-        status: member.status,
-        incarnationNumber: member.incarnationNumber,
-        piggybackCount: 0
-    });
-};
-
 RingPop.prototype.onMemberLeave = function onMemberLeave(member) {
     this.stat('increment', 'membership-update.leave');
     this.logger.debug('member has left', {
@@ -589,19 +573,16 @@ RingPop.prototype.onMembershipUpdated = function onMembershipUpdated(updates) {
     var ringChanged = false;
 
     updates.forEach(function(update) {
-        if (update.type === 'alive') {
+        if (update.status === 'alive') {
             self.onMemberAlive(update);
             ringChanged = membershipChanged = true;
-        } else if (update.type === 'faulty') {
+        } else if (update.status === 'faulty') {
             self.onMemberFaulty(update);
             ringChanged = membershipChanged = true;
-        } else if (update.type === 'leave') {
+        } else if (update.status === 'leave') {
             self.onMemberLeave(update);
             ringChanged = membershipChanged = true;
-        } else if (update.type === 'new') {
-            self.onMemberJoined(update);
-            ringChanged = membershipChanged = true;
-        } else if (update.type === 'suspect') {
+        } else if (update.status === 'suspect') {
             self.onMemberSuspect(update);
             membershipChanged = true;
         }
--- a/lib/members.js
+++ b/lib/members.js
@@ -153,23 +153,23 @@ Membership.evalOverride = function evalOverride(member, change) {
         // it affirms its "aliveness" and bumps its incarnation number.
         member.status = 'alive';
         member.incarnationNumber = +new Date();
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -531,23 +531,6 @@ RingPop.prototype.onMemberFaulty = function onMemberFaulty(member) {
     this.suspicion.stop(member);
 };
 
-RingPop.prototype.onMemberJoined = function onMemberJoined(member) {
```
jwolski: having this quasi-member state / update type just made things more confusing

lib/members.js
```diff
@@ -234,8 +234,10 @@ Membership.prototype.addMember = function addMember(member, force, noEvent) {
     this.members.splice(this.getJoinPosition(), 0, newMember);
 
     if (!noEvent) {
-        this._emitUpdated(_.extend({ type: 'new' }, newMember));
+        this._emitUpdated(_.extend({ type: newMember.status }, newMember));
```
jwolski: Don't assume 'new' or 'alive'

lib/members.js
```diff
@@ -153,23 +153,23 @@ Membership.evalOverride = function evalOverride(member, change) {
         // it affirms its "aliveness" and bumps its incarnation number.
         member.status = 'alive';
         member.incarnationNumber = +new Date();
-        return _.extend({ type: 'alive' }, member);
+        return member;
     } else if (Membership.isAliveOverride(member, change)) {
         member.status = 'alive';
         member.incarnationNumber = change.incarnationNumber || member.incarnationNumber;
-        return _.extend({ type: 'alive' }, member);
```
Raynos: so we are getting rid of the type field because its a duplicate with `status` ?

lib/members.js
```diff
@@ -234,8 +234,10 @@ Membership.prototype.addMember = function addMember(member, force, noEvent) {
     this.members.splice(this.getJoinPosition(), 0, newMember);
 
     if (!noEvent) {
-        this._emitUpdated(_.extend({ type: 'new' }, newMember));
+        this._emitUpdated(newMember);
```
Raynos: should the `status` field above be set to `'new'` ?

lib/members.js
jwolski: that's correct. `type` was just transient to communicate what type of update it was. Since I got rid of `new`, there's no longer a need for it. And `new` and `alive` are treated same.

lib/members.js
jwolski: `new` is treated the same as `alive`.

index.js
```diff
@@ -589,19 +572,16 @@ RingPop.prototype.onMembershipUpdated = function onMembershipUpdated(updates) {
...
             ringChanged = membershipChanged = true;
-        } else if (update.type === 'faulty') {
+        } else if (update.status === 'faulty') {
             self.onMemberFaulty(update);
             ringChanged = membershipChanged = true;
-        } else if (update.type === 'leave') {
+        } else if (update.status === 'leave') {
             self.onMemberLeave(update);
             ringChanged = membershipChanged = true;
-        } else if (update.type === 'new') {
```
Raynos: You are removing the `new` state, I see.

index.js
```diff
@@ -531,23 +531,6 @@ RingPop.prototype.onMemberFaulty = function onMemberFaulty(member) {
...
 };
 
-RingPop.prototype.onMemberJoined = function onMemberJoined(member) {
-    this.stat('increment', 'membership-update.new');
-    this.logger.debug('member has joined', {
-        local: this.membership.localMember.address,
-        joined: member.address
-    });
-
-    this.ring.addServer(member.address);
```
Raynos: So basically; when a member joins we should not add it to the server.

We should add it when we hear from it on the first ping or something ?

lib/members.js
Raynos: It looks like defaulting status to `alive` above is confusing.

index.js
jwolski: when a "new" member is added to the membership list, we need to use the status of that member determine how to modify the ring/membership list. Prior to this change, `new` and `alive` were treated the exact same.

lib/members.js
```diff
@@ -426,22 +428,27 @@ Membership.prototype.update = function update(changes) {
...
                 status: change.status,
                 incarnationNumber: change.incarnationNumber
-            };
-            this.addMember(member, true, true);
-            updates.push(_.extend(member, { type: 'new' }));
+            }, true, true);
+
+            updates.push(member);
+            continue;
+        }
```
Raynos: suggestion: having an `else` block might be more readable then a `continue;`

robskillington:
@jwolski is there any possibility onMembershipUpdated would get called multiple times with 'alive' status?  Seems there is no `if (!this.ring.contains(member)) { this.ring.addServer(...) }` guards

Raynos:
lgtm. However lets not default status to `'alive'` for `newMember`.

lib/members.js
jwolski: good point. Let's be more explicit! Fixed.

lib/members.js
jwolski: yes. good to point out... but im going to punt

jwolski:
@robskillington good question. the assumption is implicit that there'll only ever be 1 update per member  for calls to onMembershipUpdated. That's guaranteed by the Dissemination class (in `lib/members.js`). And you're right that there's no protection around `this.ring.contains(member)`, but the rbtree itself protects from duplicating nodes.

Raynos:
lgtm.

sunweik:
lgtm

PR #44, opened 2015-03-06
support tchannel v2

Raynos:
TChannel v2 only accepts strings and buffers. We need to
JSON stringify outside of tchannel.

reviewers: @jcorbin @jwolski


jwolski:
thanks @Raynos

PR #51, opened 2015-03-07
JSHint the lib/ directory

jwolski:
There's still some work to do to get all files in the repo to jshint correctly. I'll start with lib/.

@Raynos

### DIFF
--- a/lib/members.js
+++ b/lib/members.js
@@ -64,7 +64,10 @@ Dissemination.prototype.adjustMaxPiggybackCount = function adjustMaxPiggybackCou
 Dissemination.prototype.getChanges = function getChanges(checksum, source) {
     var changesToDisseminate = [];
 
-    for (var address in this.changes) {
+    var changedNodes = Object.keys(this.changes);
+
+    for (var i = 0; i < changedNodes.length; i++) {
+        var address = changedNodes[i];
         var change = this.changes[address];
 
         // TODO We're bumping the piggyback count even though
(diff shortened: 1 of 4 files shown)


lib/rbtree.js
```diff
@@ -188,6 +188,7 @@ RBTree.prototype.remove = function(val) {
             } else if (!isRed(node.getChild(!dir))) {
                 var sibling = p.getChild(!lastDir);
                 if (sibling !== null) {
+                    /* jshint -W073 */
```
jwolski: I chose to "ignore" instead of shuffling the code around and inserting `continue` for the sake of readability.

lib/members.js
```diff
@@ -64,7 +64,10 @@ Dissemination.prototype.adjustMaxPiggybackCount = function adjustMaxPiggybackCou
 Dissemination.prototype.getChanges = function getChanges(checksum, source) {
     var changesToDisseminate = [];
 
-    for (var address in this.changes) {
```
Raynos: :+1: a `for (var a in b)` loop forces V8 to never optimize the function.

lib/rbtree.js
```diff
@@ -40,7 +40,7 @@ RingNode.prototype.setChild = function(dir, val) {
     }
 };
 
-RBTree = function RBTree() {
+var RBTree = function RBTree() {
```
Raynos: global leaks are scary :( nice catch.

lib/rbtree.js
Raynos: i dont know what W073 is ... But ignore seems fine.

Raynos:
lgtm. Great start.

Long term we should consider moving to lint-trap.

PR #52, opened 2015-03-09
Always pass head object when emitting 'request' event

markyen:
@jwolski @Raynos 

This is done so that when a request is handled locally, the downstream handler is able to see and use the keys.

I could also duplicate the keys in both headers in order to avoid a breaking change, if we think that's better.

### DIFF
--- a/index.js
+++ b/index.js
@@ -39,6 +39,7 @@ var PingReqSender = require('./lib/swim').PingReqSender;
 var PingSender = require('./lib/swim').PingSender;
 var safeParse = require('./lib/util').safeParse;
 var RequestProxy = require('./lib/request-proxy');
+var serializeHead = require('./lib/request-proxy-util.js').serializeHead;
 var Suspicion = require('./lib/swim.js').Suspicion;
 
 var HOST_PORT_PATTERN = /^(\d+.\d+.\d+.\d+):\d+$/;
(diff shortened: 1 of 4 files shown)


jwolski:
is this necessary? are you unable to use `head` that is emitted as the third argument to `request` event listeners?

markyen:
That is a fantastic idea. Updated the PR

index.js
```diff
@@ -849,24 +850,29 @@ RingPop.prototype.handleOrProxyAll =
...
 
         dests.forEach(function(dest) {
+            var destKeys = keysByDest[dest];
             var res = hammock.Response(function(err, resp) {
                 onResponse(err, resp, dest);
             });
             if (whoami === dest) {
                 self.logger.trace('handleOrProxyAll was handled', {
-                    keys: keys,
+                    keys: destKeys,
```
Raynos: is this a seperate bugfix ?

Raynos:
lgtm.

index.js
```diff
@@ -849,24 +850,29 @@ RingPop.prototype.handleOrProxyAll =
...
                     url: req && req.url,
                     dest: dest
                 });
-                self.emit('request', req, res);
+                var head = serializeHead(req, {
```
jwolski: you'll be unnecessarily stringifying head and then having to parse again. Port what I did in this PR over to yours and you can use `rawHead`: https://github.com/uber/ringpop/pull/41/files#diff-d89c23da890a7f6cf4d117b1f209e9dcR22

index.js
markyen: It wasn't a bug in the strictest sense of the word, but it's certainly more useful now than it was before

index.js
markyen: Done

jwolski:
lgtm

Issue #36, opened 2015-02-28
benchmark convergence time and other metrics


mrhooray:
It would be good to benchmark convergence time and other metrics that we care.

It may not necessarily reflect the real situation in production but it can be used to have a rough idea on performance characteristics and as a baseline to prevent regression under difference scenarios that we can simulate.

The benchmark should be automated under different ringpop/swim configurations and scenarios like single node failure, catastrophic failure, parallel restart, rolling restart, capacity increase, packet loss, latency increase, etc.

I made a simple prototype under https://github.com/mrhooray/swim-js/tree/master/bench which measures convergence time under single node failure scenario. It does not include much automation but can be tuned with cli arguments and produces result like
```
configuration:
- cycles 20
- workers 30
- codec msgpack
- dissemination factor 15
- interval 20 ms
- join timeout 100 ms
- ping timeout 4 ms
- ping req timeout 12 ms
- ping req group size 3
- max dgram size 512 bytes
convergence time under single node failure
histogram data:
- count 20
- min 94
- max 198
- mean 130.4
- median 126
- variance 807.8315789473684
- std dev 28.422378136731773
- p75 143.75
- p95 196.74999999999997
- p99 198
```

Is this something we'd like to have? suggestions/ideas? @jwolski @Raynos

jwolski:
Yes. This is much desired.

jwolski:
this has been merged

PR #53, opened 2015-03-10
extract components with opts and default values

mrhooray:
related to #36, first step of changes that
- expose tunable parameters at initialization & runtime in the future
- unblock benchmark on convergence time with custom configuration
- separate components for easier navigation and future work


jwolski:
Looks legit, but now we have `lib/swim` dir and `lib/swim.js` which might get confusing. Thoughts about moving `lib/swim.js` to `lib/swim/index.js`?

Raynos:
:+1: more folders.

mrhooray:
@jwolski updated pr with all remaining parts of `lib/swim.js` extracted and expose suspicionTimeout on ringpop creation

index.js
```diff
@@ -106,8 +107,14 @@ function RingPop(options) {
...
     this.memberIterator = new MemberIterator(this);
-    this.gossip = new Gossip(this);
-    this.suspicion = new Suspicion(this);
+    this.gossip = new Gossip({
+        ringpop: this,
+        minProtocolPeriod: options.minProtocolPeriod
+    });
+    this.suspicion = new Suspicion({
+        ringpop: this,
+        minProtocolPeriod: options.suspicionTimeout
```
jwolski: this seems wrong. why does Suspicion need minProtocolPeriod?

mrhooray: good catch, will update

test/swim_test.js
```diff
@@ -181,6 +187,8 @@ test('marks member faulty after suspect period', function t(assert) {
 
     suspicion.start(member);
 
-    suspicion.stopAll();
-    assert.end();
+    setTimeout(function verify() {
```
jwolski: eh?

mrhooray: mocking out setTimeout used by suspicion not only does not return a timeout object, also behaves differently from code running in prod

by passing suspicionTimeout in test, suspicion would behave close to prod and this test still runs deterministically

jwolski:
looking good. you'll need a rebase now with changes. i want to spend some more time on that timeout object bug you're referencing too.

PR #56, opened 2015-03-12
benchmark convergence time

mrhooray:
benchmark convergence time with child process & message passing
future work includes
- replacing interprocess messaging with swim/ringpop cli over tchannel
- worker publish checksum on membership update to independent channel instead of interprocess messaging and master/admin watch for divergence & convergence to compute convergence time; which may apply to production as well

will parameterize benchmark after #53 and other cleanup work

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,6 +48,8 @@ var rawHead = require('./lib/request-proxy/util.js').rawHead;
 var RequestProxy = require('./lib/request-proxy/index.js');
 var safeParse = require('./lib/util').safeParse;
 var sendJoin = require('./lib/swim/join-sender.js').joinCluster;
+var adminLeaveHandler = require('./server/admin-leave-handler.js');
+var adminJoinHandler = require('./server/admin-join-handler.js');
 
 var HOST_PORT_PATTERN = /^(\d+.\d+.\d+.\d+):\d+$/;
 var MAX_JOIN_DURATION = 300000;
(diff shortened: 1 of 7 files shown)


mrhooray:
- rebased on latest master
- reintroduced leave function to ringpop; leave was refactored away, but it seems reasonable to expose leave interface for ringpop consumer
- parameterized scenario as runner argument
- added half cluster failure

```
node benchmarks/convergence-time/scenario-runner.js -h

  Usage: scenario-runner [options]

  Options:

    -h, --help          output usage information
    --scenario <value>  path to scenario relative to scenario runner
    --cycles [value]    number of cycles
    --workers [value]   number of workers
```

charliezhang:
could you rebase and merge this?

mrhooray:
cc @charliezhang, under default config
```
➜  ringpop git:(bench) node benchmarks/convergence-time/scenario-runner.js --scenario scenarios/half-cluster-failure.js
configuration:
- scenario scenarios/half-cluster-failure.js
- cycles 10
- workers 10
convergence time under half cluster failure
histogram data:
- count 10
- min 601
- max 1228
- mean 890.2
- median 893
- variance 35636.177777777775
- std dev 188.77546921615047
- p75 1030.25
- p95 1228
- p99 1228
```

index.js
```diff
@@ -190,6 +192,18 @@ RingPop.prototype.setupChannel = function setupChannel() {
     createServer(this, this.channel);
 };
 
+RingPop.prototype.adminJoin = function adminJoin(callback) {
```
jwolski: Can you remove this from the RingPop prototype? We're trying to reduce the overall footprint of the "public" RingPop interface. These two shouldn't be here.

PR #57, opened 2015-03-13
Compute a different checksum for the ring and use it for request proxying

### LINKED ISSUE #55: Maintain 2 checksums: membership and ring
The reason this is being proposed is because proxied requests are refused when membership checksums differ even if the underlying server/replicas in the ring are the same. That's because member status and incarnation number are part of the checksum. This results in more retries for proxied requests than desired.

When we validate source/destination checksums on proxy, it is to make sure that hashring lookups are consistent. And they would be if member status like suspect was not part of the checksum.

We rely heavily on membership checksums for full-syncs, etc. So we'd need 2 different checkums one for membership and the other for ring.

jwolski:
We currently compare checksums before serving a proxied request. This is generally good practice because you don't want either ends of the request to disagree on the state of the ring. But, in fact, we are comparing the consistency of membership checksums which include status and incarnation number values which really should have no impact on ring lookups other than if one of the nodes were actually removed from or added to the ring. Therefore, we're introducing a ring checksum too. This should prevent a lot of unnecessary invalid checksum errors when the ring shakes up a bit because of suspect statuses

Fixes #55

@mranney @Raynos @mrhooray

### DIFF
--- a/lib/request-proxy/send.js
+++ b/lib/request-proxy/send.js
@@ -27,7 +27,7 @@ var TypedError = require('error/typed');
 
 var MaxRetriesExceeded = TypedError({
     type: 'ringpop.request-proxy.max-retries-exceeded',
-    message: 'Max number of retries exceeded. ({maxRetries}) were attempted',
+    message: 'Max number of retries exceeded. {maxRetries} were attempted',
     maxRetries: null
 });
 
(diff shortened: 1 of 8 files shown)


lib/request-proxy/send.js
```diff
@@ -27,7 +27,7 @@ var TypedError = require('error/typed');
 
 var MaxRetriesExceeded = TypedError({
     type: 'ringpop.request-proxy.max-retries-exceeded',
-    message: 'Max number of retries exceeded. ({maxRetries}) were attempted',
+    message: 'Max number of retries exceeded. {maxRetries} were attempted',
```
jwolski: Incidental cleanup

Raynos:
lgtm.

PR #58, opened 2015-03-16
Combine integration test struct into themes

jwolski:
This cleans up the integration test directory a bit. No need to review code, just structure. I've broken the tests up into integration themes: proxying, join (maybe eventually rolled into SWIM integration), ring tests.

```
test
test/integration/
test/integration/join-test.js
test/integration/proxying-test.js
test/integration/ring-test.js
```

@mrhooray @Raynos


test/integration/index.js
```diff
@@ -1,26 +0,0 @@
-// Copyright (c) 2015 Uber Technologies, Inc.
```
Raynos: Consider keeping an `index.js` file.

There is value to being able to run `node test/integration/` and run just the integration tests.

Alternatively I guess you can run `tape test/integration/*.js`, I've always been a fan of "tests are just a node program" approach.

Raynos:
lgtm.

PR #59, opened 2015-03-19
Ping-req uses only ping status from unreachable member to initiate suspect period

jwolski:
This work is precursor to flap damping. Ping-req's needed to work better.

When nodeA pings nodeB and ping fails, nodeA initiates ping-req to `pingReqSize` random other members: nodeC, nodeD, nodeE. It's each of their responsibilities to ping nodeB on nodeA's behalf. This is known as an indirect ping. When nodeA pings the selected ping-req members (let's say nodeC), one of two things can happen:

1) nodeA cannot reach ping-req member nodeC
2) nodeA reaches nodeC, nodeC cannot reach nodeB.

The first of these two should not count against nodeB. This comes into play in flap damping in the following way. Let's say, nodeB is the flappy one and nodeA damp score for nodeB is nearing the damping threshold. Pings and ping-reqs to nodeB are failing. Every once and a while nodeB responds. It's acting erratically. What makes nodeB not think that nodeA is the flappy one? Well, nodeB has to initiate ping-reqs when pings to nodeA fails. But when nodeB is acting erratically, it's likely that nodeB is at fault for nodeA "not responding". If any of the ping-req selected members are able to reach nodeA, then nodeB will never mark nodeA suspect nor will it ever start the suspect period and ultimately never mark the node faulty. Therefore, nodeA should not ever be flappy to nodeB.

@Raynos @mrhooray @mranney

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,8 +27,8 @@ var hammock = require('uber-hammock');
 var metrics = require('metrics');
 
 var Gossip = require('./lib/swim/gossip');
-var PingReqSender = require('./lib/swim/ping-req-sender');
 var PingSender = require('./lib/swim/ping-sender');
+var sendPingReq = require('./lib/swim/ping-req-sender.js');
 var Suspicion = require('./lib/swim/suspicion');
 
 var createRingPopTChannel = require('./lib/tchannel.js').createRingPopTChannel;
@@ -248,7 +248,9 @@ RingPop.prototype.adminLeave = function adminLeave(callback) {
     });
 };
 
-RingPop.prototype.bootstrap = function bootstrap(bootstrapFile, callback) {
+RingPop.prototype.bootstrap = function bootstrap(opts, callback) {
+    var bootstrapFile = opts.bootstrapFile || opts;
+
     if (typeof bootstrapFile === 'function') {
         callback = bootstrapFile;
         bootstrapFile = null;
@@ -469,35 +471,6 @@ RingPop.prototype.protocolPing = function protocolPing(options, callback) {
     });
 };
 
-RingPop.prototype.protocolPingReq = function protocolPingReq(options, callback) {
-    this.stat('increment', 'ping-req.recv');
-
-    var source = options.source;
-    var target = options.target;
-    var changes = options.changes;
-    var checksum = options.checksum;
-
-    this.serverRate.mark();
-    this.totalRate.mark();
-    this.membership.update(changes);
-
-    var self = this;
-    this.debugLog('ping-req send ping source=' + source + ' target=' + target, 'p');
-    var start = new Date();
-    this.sendPing(target, function (isOk, body) {
-        self.stat('timing', 'ping-req-ping', start);
-        self.debugLog('ping-req recv ping source=' + source + ' target=' + target + ' isOk=' + isOk, 'p');
-        if (isOk) {
-            self.membership.update(body.changes);
-        }
-        callback(null, {
-            changes: self.issueMembershipChanges(checksum, source),
-            pingStatus: isOk,
-            target: target
-        });
-    });
-};
-
 RingPop.prototype.lookup = function lookup(key) {
     this.stat('increment', 'lookup');
     var dest = this.ring.lookup(key + '');
@@ -668,9 +641,20 @@ RingPop.prototype.pingMemberNow = function pingMemberNow(callback) {
             return callback(new Error('destroyed whilst pinging'));
         }
 
-        start = new Date();
-        self.sendPingReq(member, function() {
-            self.stat('timing', 'ping-req', start);
+        var pingReqStartTime = new Date();
+        // TODO The pinged member's status could have changed to
+        // faulty by the time we received and processed the ping
+        // response. There are no ill effects to membership state
+        // by sending a ping-req to the faulty member (and processing
+        // the response), though it does delay the protocol period
+        // unnecessarily. We may want to bypass the ping-req here
+        // if the member's status is faulty.
+        sendPingReq({
+            ringpop: self,
+            unreachableMember: member,
+            pingReqSize: self.pingReqSize
+        }, function onPingReq() {
+            self.stat('timing', 'ping-req', pingReqStartTime);
             self.isPinging = false;
 
             callback.apply(null, Array.prototype.splice.call(arguments, 0));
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -725,49 +709,6 @@ RingPop.prototype.sendPing = function sendPing(member, callback) {
     return new PingSender(this, member, callback);
 };
 
-// TODO Exclude suspect memebers from ping-req as well?
```
jwolski: Reducing Ringpop's overall public footprint.

lib/swim/ping-req-recvr.js
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
jwolski: This is an explicit ping-req handler. I named it recvr because of its relationship to "sender".

lib/swim/ping-req-sender.js
```diff
@@ -46,18 +68,165 @@ function PingReqSender(ring, member, target, callback) {
 PingReqSender.prototype.onPingReq = function (err, res1, res2) {
     if (err) {
         this.ring.logger.warn('bad response to ping-req from ' + this.member.address + ' err=' + err.message);
-        return this.callback(true);
+        this.callback(err);
```
jwolski: Reviewing my own code...

I'll wrap this with another typed error

lib/tchannel.js
```diff
@@ -42,8 +43,8 @@ var commands = {
     '/proxy/req': 'proxyReq'
 };
 
-function RingPopTChannel(ringPop, tchannel) {
-    this.ringPop = ringPop;
+function RingPopTChannel(ringpop, tchannel) {
```
jwolski: NOTE TO REVIEWER: Adopt "proper" capitalization

lib/tchannel.js
```diff
@@ -174,18 +175,19 @@ RingPopTChannel.prototype.protocolPing = function (arg1, arg2, hostInfo, cb) {
...
 
-RingPopTChannel.prototype.protocolPingReq = function (arg1, arg2, hostInfo, cb) {
+RingPopTChannel.prototype.protocolPingReq = function protocolPingReq(arg1, arg2, hostInfo, cb) {
     var body = safeParse(arg2);
     if (body === null || !body.source || !body.target || !body.changes || !body.checksum) {
         return cb(new Error('need req body with source, target, changes, and checksum'));
     }
 
-    this.ringPop.protocolPingReq({
+    recvPingReq({
```
jwolski: NOTE TO REVIEWER: No longer calling into ringpop object to handle ping-req. There's an explicit "endpoint handler"

test/integration/join-test.js
```diff
@@ -40,6 +40,7 @@ testRingpopCluster({
     size: 1
 }, 'one node can join', function t(bootRes, cluster, assert) {
     assert.ifErr(bootRes[cluster[0].hostPort].err, 'no error occurred');
+    assert.end();
```
jwolski: Note to reviewer: These tests assumed sync behavior which was fine until I wanted to use the `testRingpopCluster` framework async.

test/integration/swim-test.js
```diff
@@ -0,0 +1,200 @@
...
+// Test helpers
+function assertNumBadStatuses(assert, res, num) {
+    var badStatuses = res.pingReqErrs.filter(function filterErr(err) {
+        return err.type === 'ringpop.ping-req.bad-ping-status';
+    });
+
+    assert.equals(badStatuses.length, num, 'correct number of bad statuses');
+}
+
+function mkNoGossip(cluster) {
```
jwolski: NOTE TO REVIEWER: These are partially stubbed integration tests :|

index.js
```diff
@@ -668,9 +641,20 @@ RingPop.prototype.pingMemberNow = function pingMemberNow(callback) {
...
+        // by sending a ping-req to the faulty member (and processing
+        // the response), though it does delay the protocol period
+        // unnecessarily. We may want to bypass the ping-req here
+        // if the member's status is faulty.
+        sendPingReq({
+            ringpop: self,
+            unreachableMember: member,
+            pingReqSize: self.pingReqSize
+        }, function onPingReq() {
+            self.stat('timing', 'ping-req', pingReqStartTime);
```
Raynos: pingReqStartTime is undefined

lib/tchannel.js
Raynos: +1

Treating ringpop as a well defined server with endpoints is going to be nicer.

test/lib/test-ringpop-cluster.js
```diff
@@ -118,10 +120,14 @@ function testRingpopCluster(opts, name, test) {
 
     tape(name, function onTest(assert) {
         var cluster = bootstrapClusterOf(opts, function onBootstrap(results) {
-            test(results, cluster, assert);
+            var oldEnd = assert.end.bind(assert);
+
+            assert.end = function newEnd() {
```
Raynos: You can also listen on the end event on assert.

This might be cleaner then monkey patching assert.end

jwolski: thanks

index.js
jwolski: eh?

Raynos: how did I not see that...

test/integration/swim-test.js
```diff
@@ -0,0 +1,198 @@
...
+    var pingReqSize = 3;
+
+    sendPingReq({
+        ringpop: ringpop,
+        unreachableMember: unreachableMember,
+        pingReqSize: pingReqSize
+    }, function onPingReq(err, res) {
+        assert.ifErr(err, 'no error occurred');
+        assertNumBadStatuses(assert, res, pingReqSize);
+        assert.equals(unreachableMember.status, 'suspect',
```
Raynos: note: It seems fragile that `unreachableMember` is a mutated object. 

This test would seem more stable if you fetched the member from `ringpop.membership` again.

test/integration/swim-test.js
```diff
@@ -0,0 +1,198 @@
...
+    size: 5,
+    tap: function tap(cluster) {
+        mkNoGossip(cluster);
+    }
+}, 'ping-req inconclusive', function t(bootRes, cluster, assert) {
+    var ringpop = cluster[0];
+    var unreachableMember = ringpop.membership.findMemberByAddress(cluster[4].hostPort);
+    var pingReqSize = 3;
+
+    ringpop.membership.members.forEach(function eachMember(member) {
```
Raynos: consider leaving a comment here about "mutating address of peers so that the outgoing pings to peers fails".

Raynos:
lgtm.

test/integration/swim-test.js
jwolski: you got it

test/integration/swim-test.js
jwolski: ok

PR #60, opened 2015-03-19
Relax constraints on local overrides. Lower bound causes abundance of fu...

jwolski:
...ll-syncs

Bug fix introduced 9.5.4

@Raynos


Raynos:
lgtm

PR #62, opened 2015-03-31
Significant refactor of membership change paths in prep for smaller

jwolski:
This is ahead of some more damping work:

1) Split up `lib/member.js` into separate files
2) Reduced overall public API of Ringpop class by removing functions and delegating to internal objects, removing redundancy
3) Added `source` for updates that are applied based on MembershipUpdateRules <-- used by damping to determine flappiness

@Raynos I don't expect a full review here. That'd be too much to ask.


Raynos:
Mostly code move. looks sane to me.

PR #67, opened 2015-04-02
Minor refactor to sending pings ahead of damping

jwolski:
As the title states, a minor refactor to sending pings ahead of damping. This is just moving code around and making the send/recv pattern more uniform across gossiping.

@Raynos


Raynos:
lgtm.

PR #68, opened 2015-04-02
Factor out protocol join from Ringpop ahead of damping

jwolski:
Another refactor ahead of damping. Trying to make send/recv pattern uniform across the board. No logic change here. Just moving code.

@Raynos

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -32,10 +32,6 @@ module.exports = {
         message: 'Expected `options.app` to be a non-empty string.\n' +
             'Must specify an app for ringpop to work.\n'
     }),
-    DenyJoinError: TypedError({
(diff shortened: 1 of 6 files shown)


lib/errors.js
```diff
@@ -32,10 +32,6 @@ module.exports = {
         message: 'Expected `options.app` to be a non-empty string.\n' +
             'Must specify an app for ringpop to work.\n'
     }),
-    DenyJoinError: TypedError({
```
Raynos: note: it might be useful to keep all errors in a file. This means that other libraries can `require('ringpop/lib/errors').DenyJoinError.type` or something.

Raynos:
lgtm.

PR #69, opened 2015-04-02
Rename join_cluster to fit sender/recvr pattern that is unfolding

jwolski:
Last of the minor refactor branches that are part damping work.

Aligning send joins with rest of operations!

@Raynos


Raynos:
lgtm.

PR #72, opened 2015-04-16
Remove dependency on local member

jwolski:
@Raynos This is the bug you ran into. We're trying to access localMember in places where it may not have been added yet (aka not bootstrapping). We can just as easily access ringpop's hostPort for the same value.

More validation needs to be added around what things should be possible if you've not yet bootstrapped ringpop, but i'll leave that for another PR.

I also rewrote most of the suspicion tests in this PR because they were originally poorly written.


Raynos:
lgtm.

PR #73, opened 2015-04-16
Lookup time

jwolski:
@Raynos Allow ringpop clients to record lookup times. Moving away from the model where ringpop has statsd dependency.


Raynos:
lgtm.

PR #74, opened 2015-04-16
Reduce footprint of Ringpop's index.js by extracting endpoint handlers

jwolski:
Continue digging out all the things that should not be hanging off the Ringpop prototype.

@Raynos


Raynos:
lgtm.

PR #76, opened 2015-04-30
Thrift

jwolski:
This is no longer a work in progress and is ready for general review. I suspect that there may be further updates to the Ringpop Thrift definition, but this is good enough for some eyes.

As a part of this effort, I attempted to clean up the gossip protocol by removing some fields that were exchanged but unnecessary. That is why you may see more churn that you were expecting. The bulk of the actual Thrift integration centers around `lib/tchannel.js`
@kriskowal @leizha @Raynos @mrhooray

### DIFF
--- a/ringpop.thrift
+++ b/ringpop.thrift
@@ -1,0 +1,56 @@
+struct Change {
+    1: required string source;
+    2: required string address;
+    3: required string status;
+    4: required double incarnationNumber;
+}
+
+struct Ack {
+    1: optional list<Change> changes;
+}
+
+struct DenyingJoinsError {
+    1: required string type;
+    2: required string message;
+}
+
+struct InvalidJoinAppError {
+    1: required string type;
+    2: required string message;
+}
+
+struct InvalidJoinSourceError {
+    1: required string type;
+    2: required string message;
+}
+
+struct PingReqTargetUnreachableError {
+    1: required string type;
+    2: required string message;
+    3: optional list<Change> changes;
+}
+
+service Ringpop {
+    Ack join(
+        1: required string app,
+        2: required string source,
+        3: required double incarnationNumber
+    ) throws (
+        1: DenyingJoinsError denyingJoins,
+        2: InvalidJoinAppError invalidJoinApp,
+        3: InvalidJoinSourceError invalidJoinSource
+    );
+
+    Ack ping(
+        1: required double checksum,
+        2: optional list<Change> changes
+    );
+
+    Ack pingReq(
+        1: required double checksum,
+        2: optional list<Change> changes,
+        3: required string target
+    ) throws (
+        1: PingReqTargetUnreachableError pingReqTargetUnreachable
+    );
+}
--- a/index.js
+++ b/index.js
@@ -179,14 +179,16 @@ RingPop.prototype.destroy = function destroy() {
         this.joiner.destroy();
     }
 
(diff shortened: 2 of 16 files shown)


ringpop.thrift
```diff
@@ -0,0 +1,25 @@
...
+struct PingReqResult {
+    1: required bool pingStatus;
+    2: required list<Change> changes;
+}
+
+service Ringpop {
+    JoinResult join(1:string app, 2:string source, 3:double incarnationNumber);
+    PingResult ping(1:double checksum, 2:list<Change> changes);
+    PingReqResult pingReq(1:double checksum, 2:list<Change> changes, 3:string target);
+}
```
mrhooray: if this is an opportunity to define network interface, do we want to further simply it?

just some thoughts
- gossip layer join/sync and ping operation may reuse the same network rpc to ping and exchange membership/updates
- add meta field to Change so that we could disseminate meta/config/app updates as well
- make responses having same structure

```
struct Change {
    1: required string source;
    2: required string address;
    3: required string status;
    4: required double incarnationNumber;
    5: required string meta;
}

struct Ack {
    1: required bool ok;
    2: required list<Change> changes;
}

service Ringpop {
    Ack ping(1:double checksum, 2:list<Change> changes);
    Ack pingReq(1:double checksum, 2:list<Change> changes, 3:string target);
}
```

jwolski: I really like the consolidation of PingResult/PingReqResult into Ack, and it works well for now (I tried it), but I reverted since I felt that it was wishful thinking that the semantics would always be the same.

I also like the consolidation of the join operation into ping. There are some slightly different semantics from the way join and ping work now:
* incarnation number is exchanged during join (and not ping)
* full-sync is performed on join (and not ping)
* app is exchanged during join (and not ping)

The first and third are easily rectified by including them in the ping interface. The second is easy too, but the more I try to bend joining around the ping, the less natural the whole refactor feels.

I'm going to give it a shot though.

mrhooray: maybe the interface can be kept same by disseminating incarnation number and app via one of the changes being piggybacked on join just like local member confirms itself on treated faulty

ringpop.thrift
```diff
@@ -0,0 +1,25 @@
...
+    1: required list<Change> changes;
+}
+
+struct PingReqResult {
+    1: required bool pingStatus;
+    2: required list<Change> changes;
+}
+
+service Ringpop {
+    JoinResult join(1:string app, 2:string source, 3:double incarnationNumber);
```
leizha: if you want, you can add `required` on each input parameter as well.

however I don't recommend `required` because if you decide to deprecate these required fields in the future, you will have to manipulate the existing fields to remove `required` which is anti-pattern in thrift.
thriftify supports only thrift binary protocol for now. `required` doesn't make any difference on the wire. it ensures the field presence during serialization/deserialization. in some other thrift protocols, `required` changes how the field is serialized which makes it even harder to upgrade with requiredness changes.

ringpop.thrift
```diff
@@ -0,0 +1,25 @@
...
+    PingResult ping(1:double checksum, 2:list<Change> changes);
```
leizha: how is `checksum` calculated? why is it a double?

equal check of double type just seems strange to me. I'm not sure how reliable it is when used across languages or platforms.

PR #77, opened 2015-05-03
WIP Gossip user data

jwolski:
This is just a quick hack to get gossiping user data working, but this is the general scope of the work that needs to be done. Relatively close to completion (implementation-wise), but not quite there for a full-blown review. This PR is more for sharing purposes right now.

The technique applied for gossiping user data is the same used for gossiping membership changes. Much like the changes, user data is 'recorded' in the dissemination component. Upon the next protocol period, the recorded user data will be piggybacked with the ping request. The user data disseminated decays over time; probabilistically user data will make its way to every node. There's no state-checking/checksumming/anti-entropy of user data at this point. 

@kriskowal @Raynos @jcorbin

### DIFF
--- a/index.js
+++ b/index.js
@@ -507,7 +507,7 @@ RingPop.prototype.pingMemberNow = function pingMemberNow(callback) {
         self.stat('timing', 'ping', start);
         if (isOk) {
             self.isPinging = false;
-            self.membership.update(body.changes);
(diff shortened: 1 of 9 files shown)


jwolski:
Closing. Twas just a proof of concept. Gossiping queries, user data, config still to come...

PR #80, opened 2015-05-19
swim/join: Abort the join process if attempts exceeded

Raynos:
We currently retry joining as hard as we can; in certain
edge cases I've seen it join 600 times in a row.

We should have a cap on it.

I've also added a fixed backoff of 100ms on every join
attempt to give the network some space to breath.

r: @jwolski @kriskowal

### DIFF
--- a/lib/swim/join-sender.js
+++ b/lib/swim/join-sender.js
@@ -39,6 +39,14 @@ var JoinDurationExceededError = TypedError({
     max: null
 });
 
+var JoinAttemptsExceededError = TypedError({
+    type: 'ringpop.join-attempts-exceeded',
+    message: 'Join attempts of `{joinAttempts}` ' +
+        'exceeded max `{maxJoinAttempts}`.\n',
+    joinAttempts: null,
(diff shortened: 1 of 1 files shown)


lib/swim/join-sender.js
```diff
@@ -266,6 +303,10 @@ JoinCluster.prototype.join = function join(callback) {
                 numNodesLeft: self.joinSize - numJoined
             });
 
+            setTimeout(reJoin, self.joinRetryDelay);
```
jwolski: clear timeout somehow?

jwolski:
no tests? *sadface*

lib/swim/join-sender.js
Raynos: `JoinCluster` does not have a `destroy` method so we cannot clear it.

Raynos:
It passes tests. I just want to move on and get autobahn working again with latest tchannel.

jwolski:
definite sad face on the lack of tests and lack of clearing timer. I think you cut a corner here. But OK fine.

PR #82, opened 2015-05-27
Dedupe multiple destinations on key lookups after failed proxy

jwolski:
Destinations were not deduped properly after a proxied request failed and thus would not be retried even if the lookup resulted in a single destination (w/ the same value)

@markyen @Raynos


markyen:
LGTM

Raynos:
lgtm.

PR #83, opened 2015-05-27
remove dynamic log messages

Raynos:
We have a few of these left and they make log analysis difficult.

r: @jwolski @rf


rf:
:+1:

jwolski:
lgtm. thanks for this.

PR #93, opened 2015-06-18
Membership update speedup

jwolski:
This is a WIP, but currently trying to speed up membership updates for large clusters, especially during the join process.


jwolski:
This was getting too big. Breaking it up in favor of smaller PRs.

PR #94, opened 2015-06-21
Have dissemination component subscribe to ringChanged to reduce times…

jwolski:
… piggyback count adjustment is evaluated

This is a minor improvement. `maxPiggybackCount` was being re-evaulated with every membership update, but really it only needs to be evaluated when ring changes. This fix will just reduce the amount of times `adjustMaxPiggybackCount` is called.


Raynos:
Lgtm.

jwolski:
:) thanks @Raynos. can always count on you for weekend "lgtms"

PR #95, opened 2015-06-21
Avoid disseminating redundant changes by filtering out change that or…

jwolski:
…iginated from source

This is another optimization. Preventing issuing / piggybacking changes (in the dissemination component) to a node that created the change in the first place.

EDIT: I'll have to improve this by being a little more strict about which changes are filtered out. Filtering on source address alone is not entirely correct and might lead to interesting race conditions. Incarnation number of the member, and not just its address, should be evaluated too.

### DIFF
--- a/lib/dissemination.js
+++ b/lib/dissemination.js
@@ -54,6 +54,10 @@ Dissemination.prototype.adjustMaxPiggybackCount = function adjustMaxPiggybackCou
     }
 };
 
+Dissemination.prototype.clearChanges = function clearChanges() {
+    this.changes = [];
+};
+
 Dissemination.prototype.fullSync = function fullSync() {
     var changes = [];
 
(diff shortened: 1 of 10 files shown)


lib/swim/ping-req-sender.js
```diff
@@ -79,7 +63,19 @@ function PingReqSender(ring, member, target, callback) {
 PingReqSender.prototype.send = function send() {
     var self = this;
 
-    this.ring.channel.send(sendOptions(this), '/protocol/ping-req', null, body(this), onSend);
+    var channelOpts = {
```
jwolski: Incidental cleanup. I hated the 'body()' and 'sendOptions()' functions.

lib/swim/ping-sender.js
```diff
@@ -58,7 +58,7 @@ PingSender.prototype.send = function send() {
         host: this.address,
         timeout: this.ring.pingTimeout
     };
-    var changes = this.ring.dissemination.issueChanges();
+    var changes = this.ring.dissemination.issueChanges(null, this.address);
```
Raynos: issueChanges is not touched in this PR.

lib/dissemination.js
```diff
@@ -88,6 +92,11 @@ Dissemination.prototype.issueChanges = function issueChanges(checksum, source) {
             change.piggybackCount = 0;
         }
 
+        // Avoid issuing a change to the node from which it originated.
+        if (change.source === source) {
```
Raynos: NOICE.

lib/swim/ping-sender.js
Raynos: never mind

Raynos:
#legit

PR #115, opened 2015-07-02
Ringpop add version to latest

Alexhauser23:
CC @jwolski 

Changes added on top of latest version of master branch: 
/admin/stats now returns additional stats: TChannel version, Ringpop version, uptime, and timestamp

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,6 +25,7 @@ var fs = require('fs');
 var globalSetTimeout = require('timers').setTimeout;
 var hammock = require('uber-hammock');
 var metrics = require('metrics');
+var packageJSON = require('./package.json');
 
 var Gossip = require('./lib/swim/gossip');
 var sendPing = require('./lib/swim/ping-sender.js');
@@ -37,6 +38,7 @@ var createMembershipUpdateListener = require('./lib/membership-update-listener.j
 var createServer = require('./server');
 var Dissemination = require('./lib/dissemination.js');
 var errors = require('./lib/errors.js');
+var getTChannelVersion = require('./lib/util.js').getTChannelVersion;
 var HashRing = require('./lib/ring');
 var Membership = require('./lib/membership.js');
 var MembershipIterator = require('./lib/membership-iterator.js');
@@ -143,6 +145,11 @@ function RingPop(options) {
 
     this.destroyed = false;
     this.joiner = null;
+
+    this.startTime = Date.now(); //used for calculating uptime
+
+    this.tchannelVersion = getTChannelVersion();
+    this.ringpopVersion = packageJSON.version;
 }
 
 require('util').inherits(RingPop, EventEmitter);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -143,6 +144,12 @@ function RingPop(options) {
 
     this.destroyed = false;
     this.joiner = null;
+
+    this.startTime = Date.now(); //used for calculating uptime
+
+    var dependencies = packageJSON.devDependencies;
+    this.tchannelVersion = dependencies.tchannel;
```
jwolski: We'll need to get the tchannel version in another way. The dev dependency won't suffice since, in production, the dev version won't be the one that is running. You can get the tchannel version like this:

```
require(path.join(path.dirname(require.resolve('tchannel')), 'package.json')).version
```

jwolski:
Looking good! I left one comment about getting the tchannel version.

index.js
```diff
@@ -143,6 +145,11 @@ function RingPop(options) {
 
     this.destroyed = false;
     this.joiner = null;
+
+    this.startTime = Date.now(); //used for calculating uptime
+
+    this.tchannelVersion = require(path.join(path.dirname(require.resolve('tchannel')), 'package.json')).version;
```
jwolski: This'll require some sort of error handling. If the path for some reason does not exist and error will be thrown. Make sure the file exists / exceptions are caught, etc.

index.js
```diff
@@ -143,6 +145,24 @@ function RingPop(options) {
...
+
+    this.startTime = Date.now(); //used for calculating uptime
+
+    this.tchannelVersion = getTchannelVersion(this);
+    this.ringpopVersion = packageJSON.version;
+}
+
+function getTchannelVersion(self)
+{
+    var tchannelPackage = require(path.join(path.dirname(require.resolve('tchannel')), 'package.json'));
```
jwolski: The problem here is that if we `require()` an invalid path/module, it'll throw. You'll need to add exception handling.

index.js
```diff
@@ -143,6 +145,24 @@ function RingPop(options) {
...
+
+    if(tchannelPackage.version !== null) {
+        return tchannelPackage.version;
+    }
+    else {
+        self.logger.error('Could not find: require(path.join(path.dirname(require.resolve(tchannel)), package.json)).version');
```
jwolski: Clean this error message up. Something like `Failed to resolve tchannel version`, {
   packageJsonPath: xxx
})

index.js
```diff
@@ -359,8 +383,16 @@ RingPop.prototype.getStats = function getStats() {
...
             totalRate: this.totalRate.printObj().m1
         },
-        ring: Object.keys(this.ring.servers)
+        ring: Object.keys(this.ring.servers),
+        version: this.ringpopVersion,
+        timestamp: timestamp,
+        uptime: uptime
     };
+
+    if(this.tchannelVersion !== null) {
```
jwolski: add a space in between `if` and `(`

PR #118, opened 2015-07-06
Thrift

jwolski:
Updated version of Thrift integration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "farmhash": "^0.2.0",
     "metrics": "^0.1.8",
     "readable-stream": "^1.0.33",
-    "thriftify": "^1.0.0-alpha14",
     "uber-hammock": "^1.0.0",
     "underscore": "^1.5.2"
   },
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "add-licence": "uber-licence",
     "check-licence": "uber-licence --dry",
     "cover": "istanbul cover --print detail --report html test/index.js",
-    "jshint": "jshint --verbose *.js lib/*.js lib/request-proxy/*.js lib/swim/*.js scripts/*.js",
+    "jshint": "jshint --verbose *.js lib/*.js lib/request-proxy/*.js lib/swim/*.js server/*.js scripts/*.js",
     "travis": "npm run cover -s && istanbul report lcov && ((cat coverage/lcov.info | coveralls) || exit 0)",
     "view-cover": "opn coverage/index.html"
   },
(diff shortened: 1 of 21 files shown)


ringpop.thrift
```diff
@@ -0,0 +1,61 @@
...
+    1: optional list<Change> changes;
+    2: optional double membershipChecksum;
+}
+
+struct DenyingJoinsError {
+    1: required string type;
+    2: required string message;
+}
+
+struct InvalidJoinAppError {
```
kriskowal: s/struct/exception/. It is an alias for struct for the time being.

server/index.js
```diff
@@ -17,9 +17,13 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-
 'use strict';
 
+var fs = require('fs');
+var path = require('path');
+
+var TChannelAsThrift = require('tchannel/as/thrift');
```
kriskowal: It is now possible to extract the .AsThrift constructor from a channel instance.

package.json
```diff
@@ -29,6 +29,7 @@
     "farmhash": "^0.2.0",
     "metrics": "^0.1.8",
     "readable-stream": "^1.0.33",
+    "thriftify": "^1.0.0-alpha14",
```
Raynos: This should not be needed.

server/index.js
Raynos: tchannel is not a prod dep. please use `channel.AsThrift`

server/index.js
```diff
@@ -42,25 +46,150 @@ var commands = {
...
+    Object.keys(thriftHandlers).forEach(function eachDef(def) {
+        tchannelAsThrift.register(tchannel, def, null,
+            thriftHandlers[def]);
+    });
+}
+
+function thriftCallback(callback) {
+    return function onCallback(err, res) {
+        if (err) {
+            callback(null, new ThriftError(err));
```
Raynos: this is not a valid thing to do. Please check `err.nameAsThrift` before creating a ThriftError.

ringpop.thrift
jwolski: thx

server/index.js
jwolski: good catch w.r.t the prod dep. will use AsThrift.

package.json
jwolski: thx

server/index.js
jwolski: got it. will fix.

PR #122, opened 2015-07-15
Add ID to change that is disseminated over gossip

jwolski:
This will make traceability of how changes originate, how many hops they make, and when the last time they are applied, and who the last node was that applied it.


Raynos:
lgtm.

PR #124, opened 2015-07-18
Remove call to membership.update() in ping response. It was redundant

jwolski:
Quick fix. Call to `membership.update()` when ping response is received was redundant -- already done within `ping-sender.js`. This hints at the fact that the `pingMemberNow()` hanging off of the Ringpop prototype shouldn't even exist (just adds complexity and fragmentation of logic) and folded into the logic of PingSender... which is a refactor that'll be done next.


Raynos:
lgtm.

PR #125, opened 2015-07-20
Pass joinRetryDelay and maxJoinAttempts into join

jwolski:
Not being able to pass these into `bootstrap()` is hampering the ability to diagnose cluster-level issues that occur on startup.


Raynos:
lgtm.

PR #126, opened 2015-07-21
Reduce log level of ping-req confirming unreachable member

jwolski:
Less so much logging.

@Raynos


Raynos:
-1 please revert.

This one is actually really really useful.

PR #129, opened 2015-07-23
Wieger tick cluster go

CorgiMan:
- spawn generic ringpop processes (binaries or via interpreter). e.g.
```
tickcluster 10 path/to/binary/testpop
tickcluster 10 -i node path/to/main.js
tickcluster 10 --interpreter node path/to/main.js
```
where tickcluster is an alias for `alias tickcluster='node ~/code/ringpop/scripts/tick-cluster.js'`.

- Adds grid mode view. Press z to activate

- Also removed a bug where tickcluster crashes if you request status (type 's') directly after starting the cluster.

@jwolski, @danielheller

### DIFF
--- a/scripts/tick-cluster.js
+++ b/scripts/tick-cluster.js
@@ -26,8 +26,10 @@ var color = require('cli-color');
 var farmhash = require('farmhash').hash32;
 var generateHosts = require('./generate-hosts');
 var program = require('commander');
+var fs = require('fs');
 
-var hosts, procs, procsToStart, ringPool, localIP, toSuspend, toKill, tchannel; // defined later
+var programInterpreter, programPath, procsToStart = 5;
+var hosts, procs, ringPool, localIP, toSuspend, toKill, tchannel; // defined later
 
 /* jshint maxparams: 6 */
 
@@ -87,21 +89,21 @@ function tickAll() {
     var csums = {};
 
     hostsUp().forEach(function (host, pos, list) {
-        var body;
         var start = Date.now();
         send(host, '/admin/tick', function onSend(err, res, arg2, arg3) {
-            if (err) {
-                console.log(color.red('err: ' + err.message + ' [' + host + ']'));
-                body = JSON.stringify({checksum: 'error'});
-            }
             var durMs = Date.now() - start;
             completed.push(durMs);
-            var csum = safeParse(arg3.toString()).checksum;
-            if (csums[csum] === undefined) {
-                csums[csum] = [];
+            if (err) {
+                console.log(color.red('err: ' + err.message + ' [' + host + ']'));
+            } else {
+                var csum = safeParse(arg3.toString()).checksum;
+                if (csums[csum] === undefined) {
+                    csums[csum] = [];
+                }
+                var port = host.replace(localIP + ':', '');
+                csums[csum].push(port);
             }
-            var port = host.replace(localIP + ':', '');
-            csums[csum].push(port);
+
             if (completed.length === list.length) {
                 console.log(Object.keys(csums).sort(function (a, b) { return csums[a].length - csums[b].length; }).map(function (csum) {
                     return color.blue('[' + csums[csum].join(', ') + '] ') + color.magenta(csum + ' (' + csums[csum].length + ')');
@@ -117,26 +119,25 @@ function statsAll() {
         memberships = {};
 
     hostsUp().forEach(function (host, pos, list) {
-        var body;
         var start = Date.now();
         send(host, '/admin/stats', function onSend(err, res, arg2, arg3) {
             var durMs = Date.now() - start;
             completed.push(durMs);
             if (err) {
                 console.log(color.red('err: ' + err.message + ' [' + host + ']'));
-                body = JSON.stringify({checksum: 'error', membership: 'error'});
+            } else {
+                var membership = JSON.stringify(safeParse(arg3).membership.members);
+                
+                var csum = farmhash(membership);
+                if (csums[csum] === undefined) {
+                    csums[csum] = [];
+                    memberships[csum] = membership;
+                }
+
+                var port = host.replace(localIP + ':', '');
+                csums[csum].push(port);
             }
 
-            var membership = JSON.stringify(safeParse(arg3).membership.members);
-
-            var csum = farmhash(membership);
-            if (csums[csum] === undefined) {
-                csums[csum] = [];
-                memberships[csum] = membership;
-            }
-
-            var port = host.replace(localIP + ':', '');
-            csums[csum].push(port);
             if (completed.length === list.length) {
                 console.log(Object.keys(csums).sort(function (a, b) { return csums[a].length - csums[b].length; }).map(function (csum) {
                     return color.blue('[' + csums[csum].join(', ') + '] ') + color.magenta(memberships[csum]);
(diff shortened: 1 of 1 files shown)


CorgiMan:
I made a sloppy mistake with git. Let me know if I need to squash for decency

scripts/tick-cluster.js
```diff
@@ -146,6 +147,48 @@ function statsAll() {
     });
 }
 
+function grid() {
```
jwolski: 'grid mode' should be separated out as its own PR. It's best to keep unrelated features separate so that it's easier to revert (if need be) and review.

scripts/tick-cluster.js
```diff
@@ -429,7 +487,7 @@ function killProc(count) {
         var proc = procs[rand];
         if (proc.killed === null && proc.suspended === null) {
             logMsg(proc.port, color.green('pid ' + proc.pid) + color.red(' randomly selected for death'));
-            process.kill(proc.proc.pid, 'SIGTERM');
+            process.kill(proc.proc.pid, 'SIGKILL');
```
jwolski: how does Go handle the SIGTERM?

scripts/tick-cluster.js
```diff
@@ -472,6 +529,7 @@ function main() {
     stdin.resume();
     stdin.setEncoding('utf8');
     stdin.on('data', onData);
+    setInterval(grid, 50);
```
jwolski: For completeness sake, you should `clearInterval()` upon quitting tick-cluster

scripts/tick-cluster.js
```diff
@@ -527,10 +586,14 @@ function send(host, arg1, arg2, arg3, callback) {
 
 program
     .version(require('../package.json').version)
-    .arguments('<size>')
+    .option("-i, --interpreter [interpreter]")
```
jwolski: If <program> and <size> are required params, where do we validate them? How about providing a sane default for "size" and error out if no program is specified?

scripts/tick-cluster.js
CorgiMan: I can't find it right now but it terminates the process. You can catch the signal with the os package to shut down gracefully.

We want to simulate a node failing here so I thought SIGKILL would be more appropriate.

scripts/tick-cluster.js
CorgiMan: I'll do that

scripts/tick-cluster.js
```diff
@@ -541,8 +570,14 @@ program.on('--help', function onHelp() {
...
+
+if(!fs.existsSync(programPath)) {
+    console.log('Error: program ' + programPath + ' does not exist. Check path');
+    process.exit(1);
+}
+
+if(procsToStart === 'NaN') {
+    console.log('Error: number of processes to start is not an integer');
     process.exit(1);
 }
```
CorgiMan: Checks for validity of programPath and Number of processes to start arguments

scripts/tick-cluster.js
```diff
@@ -527,10 +547,19 @@ function send(host, arg1, arg2, arg3, callback) {
...
+    .action(function onAction(path, args, options) {
+        programInterpreter = options.interpreter;
+        if (options.N)
+            procsToStart = parseInt(options.N);
+        programPath = path;
+        if (programPath[0] !== '/') {
+            programPath = './' + programPath;
+        }
+        programArgs = args;
     });
```
CorgiMan: size is an option now (default if 5)

program path is prepended with ./ in case it is a relative path. We do this because to launch a binary in the current directory we need to type "./program". Just typing "program" would not start a process. This has the added bonus that we don't look in the path for binaries which could lead to some unexpected errors.

scripts/tick-cluster.js
```diff
@@ -26,8 +26,10 @@ var color = require('cli-color');
 var farmhash = require('farmhash').hash32;
 var generateHosts = require('./generate-hosts');
 var program = require('commander');
+var fs = require('fs');
 
-var hosts, procs, procsToStart, ringPool, localIP, toSuspend, toKill, tchannel; // defined later
+var programArgs, programInterpreter, programPath, procsToStart = 5;
+var hosts, procs, ringPool, localIP, toSuspend, toKill, tchannel; // defined later
```
CorgiMan: default procsToStart = 5 (do we want to make this a seperate variable?)

CorgiMan:
Update for usage:
size is now the '-n' option (default is 5)
e.g.:
```
tickcluster -n 10 path_to_binary 
tickcluster -n 10 -i node path_to_node_program
```

scripts/tick-cluster.js
```diff
@@ -541,8 +567,14 @@ program.on('--help', function onHelp() {
...
 
-if (!procsToStart) {
-    console.log('Error: size is a required argument');
+
+if(!fs.existsSync(programPath)) {
+    console.log('Error: program ' + programPath + ' does not exist. Check path');
+    process.exit(1);
+}
+
+if(procsToStart === 'NaN') {
```
jwolski: use `isNaN(procsToStart)`

scripts/tick-cluster.js
```diff
@@ -541,8 +567,14 @@ program.on('--help', function onHelp() {
...
 
-if (!procsToStart) {
-    console.log('Error: size is a required argument');
+
+if(!fs.existsSync(programPath)) {
+    console.log('Error: program ' + programPath + ' does not exist. Check path');
+    process.exit(1);
+}
+
+if(isNaN(procsToStart)) {
```
CorgiMan: isNan, in stead of, === 'NaN'

scripts/tick-cluster.js
```diff
@@ -527,10 +544,19 @@ function send(host, arg1, arg2, arg3, callback) {
 
 program
     .version(require('../package.json').version)
-    .arguments('<size>')
+    .option('-i, --interpreter [interpreter]')
+    .option('-n [number of ringpop nodes]')
```
jwolski: `.options('-n <nodes>')`

scripts/tick-cluster.js
```diff
@@ -527,10 +544,19 @@ function send(host, arg1, arg2, arg3, callback) {
 
 program
     .version(require('../package.json').version)
-    .arguments('<size>')
+    .option('-i, --interpreter [interpreter]')
```
jwolski: `<>` instead of `[]`

scripts/tick-cluster.js
```diff
@@ -346,8 +347,24 @@ function findLocalIP() {
...
 function ClusterProc(port) {
-    var newProc = require('child_process').spawn('node', ['main.js', '--listen=' + localIP + ':' + port, '--hosts=./hosts.json']);
+    var newProc;
+
+    if (programInterpreter) {
+        newProc = require('child_process').spawn(programInterpreter, [programPath, '--listen=' + localIP + ':' + port, '--hosts=./hosts.json'].concat(programArgs));
+    } else {
+        newProc = require('child_process').spawn(programPath, ['--listen=' + localIP + ':' + port, '--hosts=./hosts.json'].concat(programArgs));
+    }
+    newProc.on('error', function(err) {
```
jwolski: Print out the host:port of the node that died when an `error` is emitted

scripts/tick-cluster.js
```diff
@@ -346,8 +347,24 @@ function findLocalIP() {
...
+        console.log(err + 'error: failed to spawn ' + programPath);
```
jwolski: print out `err.message` instead of `err`

scripts/tick-cluster.js
```diff
@@ -527,10 +544,19 @@ function send(host, arg1, arg2, arg3, callback) {
...
-    .arguments('<size>')
+    .option('-i, --interpreter [interpreter]')
+    .option('-n [number of ringpop nodes]')
+    .arguments('<program> [args...]')
     .description('tick-cluster is a tool that launches a ringpop cluster of arbitrary size')
-    .action(function onAction(size) {
-        procsToStart = +size;
+    .action(function onAction(path, args, options) {
+        programInterpreter = options.interpreter;
+        if (options.N)
```
jwolski:
```
if (options.N) {

}
```

PR #130, opened 2015-07-25
lookupN: a function that returns the 'preference list' for a given key

weikai77:
@jwolski Can you take a look when you get a chance?

### DIFF
--- a/lib/ring.js
+++ b/lib/ring.js
@@ -26,6 +26,7 @@ function HashRing(options) {
     this.options = options || {};
 
     this.replicaPoints = this.options.replicaPoints || 100;
+    this.hashFunc = this.options.hashFunc || farmhash.hash32;
 
     this.rbtree = new RBTree();
     this.servers = {};
@@ -51,7 +52,7 @@ HashRing.prototype.addServerReplicas = function addServerReplicas(server) {
     this.servers[server] = true;
 
     for (var i = 0; i < this.replicaPoints; i++) {
-        var hash = farmhash.hash32(server + i);
+        var hash = this.hashFunc(server + i);
         this.rbtree.insert(hash, server);
     }
 };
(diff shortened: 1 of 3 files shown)


jwolski:

lib/ring.js
```diff
@@ -145,4 +146,39 @@ HashRing.prototype.lookup = function lookup(str) {
...
+    var resultArray = [];
+    var resultSet = {}; // for fast dedup
+    var hash = this.hashFunc(str);
+    var iter = this.rbtree.upperBound(hash);
+    var wrappedAround = false;
+    while (resultArray.length < n) {
+        var res = iter.str();
+        if (res === null) {
+            // must not wrap around more than once
+            if (wrappedAround) {
```
jwolski: Instead of wrappedAround as a bool, how about as the original `iter.res()`. This should prevent unnecessary duplication of lookups, yea?

weikai77:
@markyen @charliezhang @jwyuan

lib/ring.js
```diff
@@ -145,4 +146,47 @@ HashRing.prototype.lookup = function lookup(str) {
...
+    var iter = this.rbtree.upperBound(hash);
+    var first = true;
+    var firstVal;
+    while (resultArray.length < n) {
+        // remember start of loop to prevent infinite loops
+        // (This is to guard against the cases where serverCount
+        // is out of sync with the rbtree (e.g., due to a bug),
+        // e.g., n === serverCount === 3 but the rbtree contains
+        // only 2 unique servers.)
+        if (first) {
```
markyen: Could you get rid of the `if (first)` by converting this to a do…while loop?
```js
do {
    ...
} while (resultArray.length < n && iter.val() !== firstVal);
```

markyen:
lgtm

jwolski:
lgtm 2. Can you squash?

PR #131, opened 2015-07-27
add grid view, press z to toggle grid view mode

CorgiMan:


CorgiMan:
The first time the checks ran on this PR they failed. I think it was a problem with a connection.

jwolski:
I think this should be a feature of ringpop-admin and added to the top command. See this PR: https://github.com/uber/ringpop-admin/pull/4

PR #133, opened 2015-07-30
add option to allow inconsistency in request proxy

charliezhang:
cc @jwolski @iproctor @markyen

### DIFF
--- a/index.js
+++ b/index.js
@@ -108,7 +108,8 @@ function RingPop(options) {
     this.requestProxy = new RequestProxy({
         ringpop: this,
         maxRetries: options.requestProxyMaxRetries,
-        retrySchedule: options.requestProxyRetrySchedule
+        retrySchedule: options.requestProxyRetrySchedule,
+        enforceConsistency: options.enforceConsistency === undefined ? true : options.enforceConsistency
     });
 
     this.ring = new HashRing();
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -108,7 +108,8 @@ function RingPop(options) {
     this.requestProxy = new RequestProxy({
         ringpop: this,
         maxRetries: options.requestProxyMaxRetries,
-        retrySchedule: options.requestProxyRetrySchedule
+        retrySchedule: options.requestProxyRetrySchedule,
+        enforceConsistency: options.enforceConsistency === undefined ? true : options.enforceConsistency
```
jwolski: Put this condition in the constructor of the RequestProxy object instead of the instantiator. If RequestProxy is ever instantiated outside of this call-site, you'd be changing expected behavior.

aka, make this look like:

```
enforceConsistency: options.enforceConsistency
```

jwolski:
1 comment left. Otherwise, looks good to me. Thanks for the contribution!

charliezhang:
addressed comments.

jwolski:
LGTM. Thanks for the contribution. I'll get a passing test build and then merge. As you may or may not know, this'll need to be "backported" to the v9.8.x line of Ringpop. I'll do that after merge.

PR #140, opened 2015-08-11
add basic trace feature to ringpop; currently support traces that output

benfleis:
to tchannel or the given logger object.

event tracers allow a live ringpop system to dynamically foward internal
events to arbitrary events, creating space for ephemeral live event
analysis and diagnostics in running clusters.

see https://github.com/uber/ringpop/issues/139

This diff is *incomplete* -- tests are not added yet, and there are several open XXX questions for @jwolski (or others, wink wink @raynos) to consider addressing.

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,6 +62,7 @@ var rawHead = require('./lib/request-proxy/util.js').rawHead;
 var RequestProxy = require('./lib/request-proxy/index.js');
 var safeParse = require('./lib/util').safeParse;
 var sendJoin = require('./lib/swim/join-sender.js').joinCluster;
+var TracerStore = require('./lib/trace/store.js');
 
 var HOST_PORT_PATTERN = /^(\d+.\d+.\d+.\d+):\d+$/;
 var MAX_JOIN_DURATION = 300000;
@@ -153,6 +154,8 @@ function RingPop(options) {
 
     createEventForwarder(this);
 
+    this.tracers = new TracerStore(this);
+
     this.clientRate = new metrics.Meter();
     this.serverRate = new metrics.Meter();
     this.totalRate = new metrics.Meter();
@@ -180,6 +183,7 @@ RingPop.prototype.destroy = function destroy() {
     this.suspicion.stopAll();
     this.membershipUpdateRollup.destroy();
     this.requestProxy.destroy();
+    this.tracers.destroy();
 
     this.clientRate.m1Rate.stop();
     this.clientRate.m5Rate.stop();
--- a/lib/trace/config.js
+++ b/lib/trace/config.js
@@ -19,22 +19,18 @@
 // THE SOFTWARE.
 'use strict';
 
-var noop = require('./noop.js');
-
-module.exports = {
-    waitForIdentified: function (opts, callback) {
-        callback(null)
-    },
-    quit: noop,
-    close: noop,
-    request: function request(/* options */) {
-        return {
-            send: function(url, head, body, cb) {
-                cb(null, {
-                    ok: true
-                });
+var eventConfigMap = {
+    'membership.checksum.update': {
+        sourcePath: ['membership', 'checksumUpdate'],
+        sinkOptsDefaults: {
+            log: {
+                message: '/trace/membership.checksum.update'
+            },
+            tchannel: {
+                endpoint: '/trace/membership.checksum.update'
             }
-        };
-    },
-    register: function register() {}
+        }
+    }
 };
+
+module.exports = eventConfigMap;
--- a/lib/trace/core.js
+++ b/lib/trace/core.js
@@ -1,0 +1,128 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+'use strict';
+
+var events = require('events');
+var util = require('util');
+
+var _ = require('underscore');
+
+var errors = require('../errors');
+
+var tracerConfigMap = require('./config');
+
+// resolves and returns a tracer config object, with
+// traceEvent, sourceEmitter, and sourceEvent bound
+function resolveEventConfig(ringpop, traceEvent) {
+    var tracerConfig = tracerConfigMap[traceEvent];
+    if (!tracerConfig) {
+        return null;
+    }
+    var sourcePath = tracerConfig.sourcePath;
+
+    // subtle but important -- if we resolve here, then the underlying object
+    // can never change. if we do latent resolution, there may be a resource
+    // leak when we try to remove listener from something that is out of our
+    // reachable scope.
+    var emitter = getIn(ringpop, sourcePath.slice(0, -1));
+    var event = _.last(sourcePath);
+
+    return emitter && _.defaults(
+        {},
+        {
+            sourceEmitter: emitter,
+            sourceEvent: event,
+            traceEvent: traceEvent
+        },
+        tracerConfig
+    );
+}
+
+function getIn(obj, path) {
+    for (var i = 0;
+         (typeof obj === 'object' || Array.isArray(obj)) && i < path.length;
+         ++i) {
+        obj = obj[path[i]];
+    }
+    return obj;
+}
+
+var tracerOptsDefaults = {
+    expiresIn: 60000, // ms
+    // expiresAt: Date.now() + 60000,
+    sink: {type: 'log'}
+};
+
+// base for sink-specific tracers:
+// A Tracer conjoins an internal node event and a user-defined sink,
+// automatically forwarding data.
+function Tracer(ringpop, config, opts) {
+    if (!(this instanceof Tracer)) {
+        return new Tracer(opts);
+    }
+
+    this.ringpop = ringpop;
+    _.extend(this, config);
+    this.opts = _.defaults({}, opts, tracerOptsDefaults);
+    this.listener = null;
+}
+util.inherits(Tracer, events.EventEmitter);
+
+Tracer.invalidateOptions = function invalidateOptions(opts) {
+    if (!opts) {
+        return errors.InvalidOptionError({option: 'opts', reason: 'is missing'});
+    }
+    if ((opts.expiresIn && opts.expiresAt) ||
+        !(opts.expiresIn || opts.expiresAt)) {
+        return errors.InvalidOptionError({
+            option: 'expires',
+            reason: 'must specify exactly one of expiresAt, expiresIn'});
+    }
+    if (!opts.sink) {
+        return errors.InvalidOptionError({option: 'sink', reason: 'is missing'});
+    }
+    return false;
+};
+
+Tracer.prototype.invalidate = function invalidate() {
+    return Tracer.invalidateOptions(this.opts);
+};
+
+Tracer.prototype.connectSource = function connectSource() {
+    if (!this.listener) {
+        this.listener = this.send.bind(this);
+        this.sourceEmitter.on(this.sourceEvent, this.listener);
+    }
+    process.nextTick(this.emit.bind(this, 'connectSource'));
+};
+
+Tracer.prototype.disconnectSource = function disconnectSource() {
+    if (this.listener) {
+        this.sourceEmitter.removeListener(this.sourceEvent, this.listener);
+    }
+    this.listener = null;
+    process.nextTick(this.emit.bind(this, 'disconnectSource'));
+};
+
+module.exports = {
+    resolveEventConfig: resolveEventConfig,
+    Tracer: Tracer
+};
+
--- a/lib/trace/index.js
+++ b/lib/trace/index.js
@@ -17,24 +17,8 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
+
 'use strict';
 
-var noop = require('./noop.js');
+module.exports = require('./store');
 
-module.exports = {
-    waitForIdentified: function (opts, callback) {
-        callback(null)
-    },
-    quit: noop,
-    close: noop,
-    request: function request(/* options */) {
-        return {
-            send: function(url, head, body, cb) {
-                cb(null, {
-                    ok: true
-                });
-            }
-        };
-    },
-    register: function register() {}
-};
--- a/lib/trace/log.js
+++ b/lib/trace/log.js
@@ -1,0 +1,131 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+'use strict';
+
+var _ = require('underscore');
+var util = require('util');
+
+var errors = require('../errors');
+
+var core = require('./core');
+
+//
+// Log forwarder
+//
+// opts = {
+//     sink: {
+//         // REQUIRED
+//         type: "log"
+//
+//         // OPTIONAL/HAS DEFAULTS
+//         level: 'info', // from [error,warn,info,debug,trace]
+//         message: 'trace'
+//     }
+// }
+//
+
+var logOptsDefaults = {
+    sink: {
+        type: 'log',
+        level: 'info',
+        message: 'trace'
+    }
+};
+
+var logLevels = {
+    error: true,
+    warn: true,
+    info: true,
+    debug: true,
+    trace: true
+};
+
+function LogTracer(ringpop, config, opts) {
+    if (!(this instanceof LogTracer)) {
+        return new LogTracer(opts);
+    }
+
+    opts.sink = _.defaults(
+        {},
+        opts.sink,
+        config.sinkOptsDefaults.log,
+        logOptsDefaults.sink
+    );
+    core.Tracer.call(this, ringpop, config, opts);
+
+    this.level = this.opts.sink.level;
+    this.message = opts.sink.message || null;
+}
+util.inherits(LogTracer, core.Tracer);
+
+
+LogTracer.prototype.invalidate = function invalidate() {
+    var sinkOpts = this.opts.sink;
+    var err = core.Tracer.prototype.invalidate.call(this);
+
+    if (err) {
+        return err;
+    }
+    if (!logLevels[sinkOpts.level]) {
+        return errors.InvalidOptionError({option: 'log.level', reason: 'is missing'});
+    }
+    return false;
+};
+
+LogTracer.prototype.connectSink = function connectSink(callback) {
+    if (callback) {
+        this.once('connectSink', callback);
+    }
+    process.nextTick(this.emit.bind(this, 'connectSink'));
+};
+
+LogTracer.prototype.disconnectSink = function disconnectSink(callback) {
+    if (callback) {
+        this.once('disconnectSink', callback);
+    }
+
+    this.disconnectSource();
+    process.nextTick(this.emit.bind(this, 'disconnectSink'));
+};
+
+LogTracer.prototype.send = function send(blob, callback) {
+    if (callback) {
+        this.once('send', callback);
+    }
+
+    if (Buffer.isBuffer(blob)) {
+        blob = blob.toString();
+    } else if (typeof blob === 'object') {
+        blob = JSON.stringify(blob);
+    }
+
+    this.ringpop.logger[this.level](this.message, blob);
+    process.nextTick(this.emit.bind(this, 'send'));
+};
+
+LogTracer.prototype.matches = function matches(traceEvent, opts) {
+    return (
+        this.traceEvent === traceEvent &&
+        opts.sink.type === 'log' &&
+        opts.sink.message === this.opts.sink.message);
+};
+
+
+module.exports = LogTracer;
--- a/lib/trace/store.js
+++ b/lib/trace/store.js
@@ -1,0 +1,165 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+//
+// A lightweight store for managing tracers.
+//
+
+var EventEmitter = require('events').EventEmitter;
+
+var _ = require('underscore');
+
+var core = require('./core');
+var types = require('./types');
+
+function TracerStore(ringpop, opts) {
+    opts = opts || {};
+    opts.timers = opts.timers || {};
+
+    if (!(this instanceof TracerStore)) {
+        return new TracerStore(ringpop);
+    }
+    this.ringpop = ringpop;
+    this.tracers = {};
+    this.timers = {
+        nextTick: opts.timers.nextTick || process.nextTick,
+        setTimeout: opts.timers.setTimeout || ringpop.setTimeout || setTimeout,
+        clearTimeout: opts.timers.clearTimeout || clearTimeout,
+        now: opts.timers.now || Date.now
+    };
+}
+require('util').inherits(TracerStore, EventEmitter);
+
+TracerStore.prototype.add = function add(config, opts, callback) {
+    var self = this;
+    var tracerOptsStr = JSON.stringify(opts);
+
+    if (callback) {
+        self.once('add', callback);
+    }
+
+    var err = core.Tracer.invalidateOptions(opts);
+    if (err) {
+        process.nextTick(self.emit.bind(self, 'add', err));
+        return null;
+    }
+
+    var tracer = self._get(config, opts);
+    if (!tracer) {
+        var TracerType = types[opts.sink.type];
+        tracer = new TracerType(self.ringpop, config, opts);
+        err = tracer.invalidate(opts);
+        if (err) {
+            process.nextTick(self.emit.bind(self, 'add', err));
+            return null;
+        }
+
+        var eventTracers = self.tracers[config.traceEvent];
+        if (!eventTracers) {
+            eventTracers = [];
+            self.tracers[config.traceEvent] = eventTracers;
+        }
+        eventTracers.push(tracer);
+
+        tracer.connectSink(function onConnect(err) {
+            if (err) {
+                self.remove(config, opts, null);
+                self.emit('add', err, null);
+                return;
+            }
+            tracer.connectSource();
+            self._updateExpiration(config, opts, tracer);
+            process.nextTick(self.emit.bind(self, 'add', null, tracerOptsStr));
+        });
+    } else {
+        self._updateExpiration(config, opts, tracer);
+        process.nextTick(self.emit.bind(self, 'add', null, tracerOptsStr));
+    }
+
+    return tracer;
+};
+
+TracerStore.prototype.remove = function remove(config, opts, callback) {
+    if (callback) {
+        this.once('remove', callback);
+    }
+
+    var tracer = null;
+    var i = this._indexOf(config, opts);
+    if (i >= 0) {
+        tracer = this.tracers[config.traceEvent][i];
+    }
+    if (tracer) {
+        if (tracer._timer) {
+            this.timers.clearTimeout(tracer._timer);
+            tracer._timer = null;
+        }
+        this.tracers[config.traceEvent].splice(i, 1);
+        tracer.disconnectSink();
+        tracer.disconnectSource();
+    }
+    this.emit('remove', null, tracer);
+    return tracer;
+};
+
+TracerStore.prototype._indexOf = function _indexOf(config, opts) {
+    var traceEvent = config.traceEvent;
+    var tracers = this.tracers[traceEvent] || [];
+    for (var i = 0; i < tracers.length; ++i) {
+        var tracer = tracers[i];
+        if (tracer.matches(traceEvent, opts)) {
+            return i;
+        }
+    }
+    return -1;
+};
+
+TracerStore.prototype._get = function _get(config, opts) {
+    var i = this._indexOf(config, opts);
+    return i >= 0 ? this.tracers[config.traceEvent][i] : null;
+};
+
+TracerStore.prototype._updateExpiration = function _updateExpiration(config, opts, tracer) {
+    var now = this.timers.now();
+    var newOffset = Math.max(0, opts.expiresIn || (opts.expiresAt - now));
+    var removeFn = this.remove.bind(this, config, opts);
+
+    if (tracer._timer) {
+        this.timers.clearTimeout(tracer._timer);
+        tracer._timer = null;
+    }
+
+    // 0/null offset implies already expired, thus don't re-add
+    if (newOffset) {
+        tracer._timer = this.timers.setTimeout(removeFn, newOffset);
+    }
+};
+
+TracerStore.prototype.destroy = function destroy() {
+    var self = this;
+    _.each(self.tracers, function removeTracer(tracer) {
+        self.remove(tracer.config, tracer.opts);
+    });
+    self.emit('destroyed');
+};
+
+module.exports = TracerStore;
--- a/lib/trace/tchannel.js
+++ b/lib/trace/tchannel.js
@@ -1,0 +1,176 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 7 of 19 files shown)


lib/trace/index.js
```diff
@@ -0,0 +1,156 @@
...
+    // XXX testing only?
+    //
+    // @wolski, we chatted briefly about this -- whatcha think should
+    // be done? the testing framework doesn't work, as is. (I can add
+    // more mocks to make it go.)
+    if (!ringpop || !ringpop.channel || !ringpop.membership) {
+        return;
+    }
+
+    registerTraceCommand(ringpop, '/trace/add', createTracer);
```
jwolski: It would be more conventional if you registered these endpoint handlers in server/index.js and created handlers trace-add-handler.js and trace-remove-handler.js under server/

lib/trace/index.js
```diff
@@ -0,0 +1,156 @@
...
+        case 'tchannel':
+            tracer = new TChannelTracer(ringpop, config, opts);
+            break;
+
+        case 'log':
+            tracer = new LogTracer(ringpop, config, opts);
+            break;
+
+        default:
+            // can't occur since validate throws first, but jic.
```
jwolski: I feel it would be simpler not to use exception throwing for control flow. How about just:

```
callback(err);
return;
```

right here?

lib/trace/index.js
```diff
@@ -0,0 +1,156 @@
...
+    callback = callback || function() {};
+
+    if (!core.Tracer.validate(opts)) {
+        throw errors.InvalidOptionError({ option: 'sink.type', reason: 'is empty or unknown' });
+    }
+
+    if (indexOfTracer(ringpop, config, opts) >= 0) {
+        // idempotent, happy noop
+        callback();
+        return;
```
jwolski: In the catch block you return `null`. Here you don't. It's inconsistent. Is there a reason?

lib/trace/index.js
```diff
@@ -0,0 +1,156 @@
...
+function registerTraceCommand(ringpop, endpoint, processor) {
+    var logPrefix = 'ringpop' + endpoint.replace(/\//g, '.');
+
+    ringpop.channel.register(endpoint, function onRequest(req, resp, optsStr) {
+        resp.headers.as = 'raw';
+        var opts = toCamelCase(safeParse(optsStr));
+        var config = core.resolveTracerConfig(ringpop, opts.event);
+        if (!config) {
+            var err = 'invalid event: ' + opts.event;
+            ringpop.logger.info(logPrefix + ' error: ' + err);
```
jwolski: Use constant strings for logger messages and a "meta" object for variable data as the second argument:

```
ringpop.logger.info('ringpop invalid event registered', {
  variableData: here
});
```

It provides better searchability especially in a tool like ELKibana

lib/trace/index.js
```diff
@@ -0,0 +1,156 @@
...
+            resp.sendNotOk(err);
+            return;
+        }
+
+        processor(ringpop, config, opts, function done(err) {
+            if (err) {
+                ringpop.logger.info(logPrefix + ' error: ' + err);
```
jwolski: how about warn instead of info? it seems more appropriate given that it is abnormal behavior. Thoughts?

lib/trace/core.js
```diff
@@ -0,0 +1,121 @@
...
+
+    this.ringpop = ringpop;
+    _.extend(this, config);
+    this.opts = opts;
+    this.listener = null;
+}
+util.inherits(Tracer, events.EventEmitter);
+
+Tracer.validate = function validate(opts) {
+    var knownTypes = { 'tchannel': true, 'log': true };
```
jwolski: This is quite hidden. Anyway you can promote `knownTypes` to be more visible?

lib/trace/core.js
```diff
@@ -0,0 +1,121 @@
...
+            tchannel: {
+                endpoint: '/trace/membership.update'
+            }
+        }
+    }
+};
+
+// resolves and returns a tracer config object, with
+// extEvent, sourceEmitter, sourceEvent
+function resolveTracerConfig(ringpop, extEvent) {
```
jwolski: extract to trace/config.js?

lib/trace/index.js
benfleis: Makes sense. this is a side effect of originally providing individual per-event endpoints.
But also, it depends a bit on how much you prefer that style where "rm -rf lib/trace" would almost just work. tell me what you prefer, and I will arrange it.

lib/trace/core.js
```diff
@@ -0,0 +1,121 @@
...
+    var eventConfig = eventConfigMap[extEvent];
+    var emitterPath = eventConfig.sourcePath;
```
jwolski: the lack of a `sourcePath` is more the fault of the ringpop developer at implementation time rather at trace add/remove time. I feel like the concerns of improper `eventConfigMap` definition should be separated from resolution tracer config. Thoughts?

lib/trace/index.js
benfleis: this is a side effect of doing type-specific validation within the constructors. I could do it explicitly instead, and then use traditional flow control. (I wrote it that way first, then went the other way.) Mostly a question of "good practice", and I don't have a notion on this particular issue. You?

lib/trace/index.js
benfleis: The inner block is a return from a callback, and thus is swallowed.

the outer code returns either the tracer, or null.

lib/trace/index.js
benfleis: beautiful! fixing.

lib/trace/index.js
benfleis: yep, mistake.

lib/trace/log.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+    process.nextTick(this.emit.bind(this, 'connectSink'));
+};
+
+LogTracer.prototype.disconnectSink = function disconnectSink(callback) {
+    if (callback) {
+        this.once('disconnectSink', callback);
+    }
+
+    self.disconnectSource();
```
jwolski: self is not defined

lib/trace/core.js
benfleis: done

lib/trace/core.js
benfleis: me likey. done.

lib/trace/core.js
benfleis: Absolutely true, and good grasp :) I wrote this as an assert/return combo to check myself, but didn't clean it properly. Gone.

lib/trace/log.js
benfleis: fixed (along with a few other dangling/now unneeded selfs)

lib/trace/index.js
jwolski: oops

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+    var tracers = ringpop.tracers[extEvent] || [];
+    for (var i = 0; i < tracers.length; ++i) {
+        var tracer = tracers[i];
+        if (tracer.matches(extEvent, opts)) {
+            return i;
+        }
+    }
+    return -1;
+}
+
```
jwolski: reduce spacing to one line (instead of 3)

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+var TChannelTracer = require('./tchannel');
+var LogTracer = require('./log');
+
+function registerTraceEndpoints(ringpop) {
+    // XXX testing only?
+    //
+    // @wolski, we chatted briefly about this -- whatcha think should
+    // be done? the testing framework doesn't work, as is. (I can add
+    // more mocks to make it go.)
+    if (!ringpop || !ringpop.channel || !ringpop.membership) {
```
jwolski: I'm not sure what exactly you're asking in your XXX. I think you're referring to the validation/existence checks of ringpop.channel and ringpop.membership. Based on that assumption, I want to move away from mocks. When we test Ringpop it should be a fully-hydrated Ringpop instance with all of its bells and whistles (channel and membership). It makes tests much easier reason to reason about when you're trying to decipher who/what is being tested.
In generaLet's move away from mocks. I think all/most Ringpop

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+        return;
+    }
+
+    registerTraceCommand(ringpop, '/trace/add', createTracer);
+    registerTraceCommand(ringpop, '/trace/remove', destroyTracer);
+}
+
+function registerTraceCommand(ringpop, endpoint, processor) {
+    var logPrefix = 'ringpop' + endpoint.replace(/\//g, '.');
```
jwolski: there's an unwritten convention for ringpop logging. as an example, it's been something like:

```
ringpop.logger.warn('ringpop tracing received invalid event', {
    local: ringpop.whoami(),
    err: err
});
```

NOTE: Without the colon, with more context `local` property.

You might think, "why don't we encode this convention in a function?". And that'd be right of you to ask.

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
jwolski: This file definitely has more of a "server" element to it. That's obvious in that you're setting up the endpoint handlers. Existing convention is for handlers to live in `server/`. Handlers shouldn't have a whole lot of meet in them, other than validating request parameters, delegating to a 'controller' (aka some module in `lib/trace`) and then crafting a tchannel response from the results of the controller.

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+    registerTraceCommand(ringpop, '/trace/remove', destroyTracer);
+}
+
+function registerTraceCommand(ringpop, endpoint, processor) {
+    var logPrefix = 'ringpop' + endpoint.replace(/\//g, '.');
+
+    ringpop.channel.register(endpoint, function onRequest(req, resp, optsStr) {
+        resp.headers.as = 'raw';
+        var opts = toCamelCase(safeParse(optsStr));
+        var config = core.resolveTracerConfig(ringpop, opts.event);
```
jwolski: I think this function should be named `resolveEventConfig`. The way it is written currently doesn't scream out to me, 'resolve Tracer Config for Event'

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+        if (!config) {
+            var err = errors.InvalidOptionError({ option: 'event', reason: 'it is invalid'});
+            ringpop.logger.warn(logPrefix + ': invalid event', err);
```
jwolski: See above remark on logging conventions.

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+            resp.sendNotOk(err);
+            return;
+        }
+
+        processor(ringpop, config, opts, function done(err) {
```
jwolski: can you call it `handler` instead of processor? It's more conventional. See *-handlers in server/

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+            if (err) {
+                ringpop.logger.warn(logPrefix + ': remote error', err);
```
jwolski: See above remark on logging conventions.

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+                resp.sendNotOk(err);
```
jwolski: See above remark on logging conventions.

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+                return;
+            } 
+            ringpop.logger.info(logPrefix + ': success', opts);
```
jwolski: See above remark on logging conventions.

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+            resp.sendOk();
+        });
+    });
+}
+
+var knownTracerTypes = {
```
jwolski: thoughts on pulling this out into `lib/trace/types.js`? It might seem like overkill, but I do think it promotes supported tracer types to a higher-level.

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+    tchannel: true,
+    log: true
+};
+
+function createTracer(ringpop, config, opts, callback) {
```
jwolski: rename to `addTracer`

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+    callback = callback || function() {};
+
+    if (!core.Tracer.validate(opts) || !knownTracerTypes[opts.sink.type]) {
```
jwolski: sink type and tracer type seem to be used interchangeably. Can you decide on a single name and then be consistent throughout?

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+        callback(errors.InvalidOptionError(
+            { option: 'sink.type', reason: 'is empty or unknown' }));
+        return;
+    }
+
+    if (indexOfTracer(ringpop, config, opts) >= 0) {
```
jwolski: Rather than the registered tracers hanging off of the ringpop instance, what are your thoughts about creating an explicit TracerStore aka `lib/trace/store.js` for registered tracers and thus you can call something like:

```
if (store.hasTracer(...)) {
  // idempotent
}
```

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+        tracer = new TChannelTracer(ringpop, config, opts);
+        break;
+
+    case 'log':
+        tracer = new LogTracer(ringpop, config, opts);
+        break;
+
+    default:
+        // can't occur since validate throws first, but jic.
+        callback(new Error('wtf batman!??! impossible sink.type detected'));
```
jwolski: No comic book characters plz ;).  Let's just do legitimate error messages.

lib/trace/core.js
```diff
@@ -0,0 +1,108 @@
...
+util.inherits(Tracer, events.EventEmitter);
+
+Tracer.validate = function validate(opts) {
+    return (
+        opts &&
+        opts.sink
+    );
+};
+
+Tracer.prototype.validate = function validate() {
```
jwolski: An instance-level alias for a class-level function is a bit redundant/unnecessary to me. Can you stick with one or the other?

lib/trace/core.js
```diff
@@ -0,0 +1,108 @@
...
+var events = require('events');
+var util = require('util');
+
+var _ = require('underscore');
+
+var tracerConfigMap = require('./config');
+
+// resolves and returns a tracer config object, with
+// extEvent, sourceEmitter, sourceEvent
+function resolveTracerConfig(ringpop, extEvent) {
```
jwolski: what does `extEvent` refer to here? To be more consistent, I'd call it `sinkEvent`, but first see my comments below about sink and naming.

lib/trace/core.js
```diff
@@ -0,0 +1,108 @@
...
+    var tracerConfig = tracerConfigMap[extEvent];
+    var emitterPath = tracerConfig.sourcePath;
+
+    // XXX subtle but important -- if we resolve here, then the
+    // underlying object can never change. if we do latent resolution,
+    // there may be a resource leak when we try to remove listener
+    // from something that is out of our reachable scope.
+    var emitter = getIn(ringpop, emitterPath.slice(0, -1));
```
jwolski: this technique of being able to attach to any event emitter is very clever and extremely convenient.

lib/trace/config.js
```diff
@@ -0,0 +1,36 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+'use strict';
+
+var eventConfigMap = {
+    'membership.update': {
```
jwolski: any reason the key to this map isn't `membership.updated` to be consistent with the actual event that is emitted?

lib/trace/core.js
```diff
@@ -0,0 +1,108 @@
...
+
+var _ = require('underscore');
+
+var tracerConfigMap = require('./config');
+
+// resolves and returns a tracer config object, with
+// extEvent, sourceEmitter, sourceEvent
+function resolveTracerConfig(ringpop, extEvent) {
+    var tracerConfig = tracerConfigMap[extEvent];
+    var emitterPath = tracerConfig.sourcePath;
```
jwolski: assigning the value of `sourcePath` to a variable of a different name is a bit confusing. it adds to the cognitive load on the reader. any reason not to call it the same thing?

```
var sourcePath = tracerConfig.sourcePath;
```

lib/trace/core.js
```diff
@@ -0,0 +1,108 @@
...
+
+    // XXX subtle but important -- if we resolve here, then the
+    // underlying object can never change. if we do latent resolution,
+    // there may be a resource leak when we try to remove listener
+    // from something that is out of our reachable scope.
+    var emitter = getIn(ringpop, emitterPath.slice(0, -1));
+    var event = _.last(emitterPath);
+
+    return _.extend(
```
jwolski: The results of the config lookup can be memoized, cant they?

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+        return null;
+    }
+    if (!tracer.validate(opts)) {
+        callback(errors.InvalidOptionError(
+            { option: 'sink', reason: 'has invalid option(s)' }));
+        return null;
+    }
+
+    var eventTracers = ringpop.tracers[config.extEvent];
```
jwolski: You're registering tracers based on 'user input' (the `extEvent` value). Even though the value of `extEvent` is validated against the tracer config map, it leaves the reader with an unsettling feeling. I think it's just in the naming... 

Change `config.extEvent` to `config.eventPath` where eventPath could equal `sourcePath.join('.')`

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+    if (!eventTracers) {
+        eventTracers = [];
+        ringpop.tracers[config.extEvent] = eventTracers;
+    }
+    eventTracers.push(tracer);
+
+    tracer.connectSink(function onConnect(err) {
+        if (err) {
+            destroyTracer(ringpop, config, opts, null);
```
jwolski: See comment above about TracerStore/TracerRegistry

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+            callback(err);
+            return;
+        }
+        tracer.connectSource(tracer);
```
jwolski: Passing itself to `connectSource()` looked strange to me. I verified that connectSource does not actually take/use a first argument.

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+        callback(null, tracer);
+    });
+
+    return tracer;
+}
+
+function destroyTracer(ringpop, config, opts, callback) {
```
jwolski: rename to removeTracer

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    var eventTracers = ringpop.tracers[config.extEvent];
+    if (!eventTracers) {
+        eventTracers = [];
+        ringpop.tracers[config.extEvent] = eventTracers;
+    }
+    eventTracers.push(tracer);
+
+    tracer.connectSink(function onConnect(err) {
```
jwolski: any reason not to roll connectSink and connectSource into one function?

lib/trace/core.js
```diff
@@ -0,0 +1,108 @@
...
+    if (this.listener) {
+        this.sourceEmitter.removeListener(this.sourceEvent, this.listener);
+    }
+    this.listener = null;
+};
+
+
+
+module.exports = {
+    resolveTracerConfig: resolveTracerConfig,
```
jwolski: move resolveTracerConfig to config.js as well? It's more logical to me to do so. And that way you wouldn't have to export the config map.

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+}
+
+function destroyTracer(ringpop, config, opts, callback) {
+    callback = callback || function() {};
+    var i = indexOfTracer(ringpop, config, opts);
+    if (i >= 0) {
+        var tracers = ringpop.tracers[config.extEvent];
+        var tracer = tracers[i];
+        tracers.splice(i, 1);
+        tracer.disconnectSink();
```
jwolski: similar to the connect() recommendation: can disconnectSink and disconnectSource be merged into one?

lib/trace/tchannel.js
```diff
@@ -0,0 +1,163 @@
...
+        };
+        this.ringpop.logger.warn(err.message);
+        this.emit(event, err);
+        return;
+    }
+    this.emit(event, null, resp);
+};
+
+TChannelTracer.prototype.matches = function matches(extEvent, opts) {
+    return (
```
jwolski: Interesting. I think matching on the option-set should be fine. Had you considered a token-exchange mechanism? In other words, generate a unique token for the tracer upon adding it and use that same token for future communication.

lib/trace/tchannel.js
```diff
@@ -0,0 +1,163 @@
...
+                'cn': 'ringpop'
+            }
+        },
+        _.omit(sinkOpts, 'type', 'hostPort')
+    );
+
+    // trigger early identification, minimize latency
+    this.ringpop.channel.waitForIdentified(
+        _.pick(this.reqOpts, 'host'),
+        this._remoteError.bind(this, 'connectSink'));
```
jwolski: _remoteError doesn't seem to capture what's going on. It handles both error and non error cases. How about simply

this.ringpop.channel.waitForIdentified(..., function onIdentified(err) {
  // handling
});

lib/trace/tchannel.js
```diff
@@ -0,0 +1,163 @@
...
+        self.emit('disconnectSink');   
+    });
+};
+
+TChannelTracer.prototype.send = function send(blob, callback) {
+    if (callback) {
+        this.once('send', callback);
+    }
+
+    if (typeof blob !== 'string') { // XXX also allow Buffer?
```
jwolski: Addresssing XXX: does it hurt to use just invoke `toString())` on `blob` in any case?

lib/trace/tchannel.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+        err.message = err.message +
+                ' [' + this.opts.hostPort + ' ' + this.reqOpts.endpoint + ']';
+        err.context = {
+            message: err.message,
+            reqOpts: this.reqOpts,
+            endpoint: this.reqOpts.endpoint,
+        };
+        this.ringpop.logger.warn(err.message);
+        this.emit(event, err);
```
jwolski: who do you envision as being the listeners of this event?

lib/trace/index.js
```diff
@@ -0,0 +1,165 @@
...
+    if (!tracer.validate(opts)) {
+        callback(errors.InvalidOptionError(
+            { option: 'sink', reason: 'has invalid option(s)' }));
+        return null;
+    }
+
+    var eventTracers = ringpop.tracers[config.extEvent];
+    if (!eventTracers) {
+        eventTracers = [];
+        ringpop.tracers[config.extEvent] = eventTracers;
```
jwolski: I don't see any mechanism to sweep and expire stale event tracers. Still to come? Or ignoring?

lib/trace/tchannel.js
```diff
@@ -0,0 +1,163 @@
...
+        this.once('disconnectSink', callback);
+    }
+
+    process.nextTick(function() {
+        self.disconnectSource();
+        self.emit('disconnectSink');   
+    });
+};
+
+TChannelTracer.prototype.send = function send(blob, callback) {
```
jwolski: `send()` acts as the event listener and 'event forwarder'. Are you assuming here that `blob` is the only argument passed to event listeners? Or am I missing something?

lib/trace/core.js
```diff
@@ -0,0 +1,108 @@
...
+    var tracerConfig = tracerConfigMap[extEvent];
+    var emitterPath = tracerConfig.sourcePath;
+
+    // XXX subtle but important -- if we resolve here, then the
+    // underlying object can never change. if we do latent resolution,
+    // there may be a resource leak when we try to remove listener
+    // from something that is out of our reachable scope.
+    var emitter = getIn(ringpop, emitterPath.slice(0, -1));
+    var event = _.last(emitterPath);
+
```
jwolski: you're very much trusting the structure of the tracerConfigMap for each event. That's fine. But write tests so that devs know when they've written an invalid event config.

jwolski:
I did lots of comments, but the overall approach looks solid. I like where this is going. Of course, this deserves all sorts of tests to be written.

lib/trace/core.js
```diff
@@ -0,0 +1,108 @@
...
+var tracerConfigMap = require('./config');
+
+// resolves and returns a tracer config object, with
+// extEvent, sourceEmitter, sourceEvent
+function resolveTracerConfig(ringpop, extEvent) {
+    var tracerConfig = tracerConfigMap[extEvent];
+    var emitterPath = tracerConfig.sourcePath;
+
+    // XXX subtle but important -- if we resolve here, then the
```
jwolski: Not 100% certain by what you mean here, but I do get the point that we are relying on some of the underlying components of Ringpop not to be reassigned. For example, if we attach listeners to `ringpop.membership` and that value is reassigned somewhere else in the codebase, we won't ever re-register or de-register previous listeners.

lib/trace/log.js
```diff
@@ -0,0 +1,123 @@
...
+    this.disconnectSource();
+    process.nextTick(this.emit.bind(self, 'disconnectSink'));
+};
+
+LogTracer.prototype.send = function send(blob, callback) {
+    if (callback) {
+        this.once('send', callback);
+    }
+
+    if (typeof blob !== 'string') { // XXX also allow Buffer?
```
jwolski: Addressing XXX: Can we just blob && blob.toString()?

lib/trace/config.js
benfleis: The current actual event doesn't emit the data that would be needed for convergence. The external will be renamed; we should probably give a few moments thought to what these should look like, since they will become advertised and external.

lib/trace/core.js
benfleis: updated to traceEvent

lib/trace/core.js
benfleis: no reason. fixed.

lib/trace/core.js
benfleis: I think it's a safe assumption too, but wanted to make note of the assumption itself. I'll kill the XXX, but leave the comment?

And as to global emitter, I briefly thought about that, and also dislike it. If you change your mind, let me know ;)

lib/trace/core.js
benfleis: Agreed. The code had error handling, but I failed to correctly return falsy if emitter resolved to null. So if an attempt to attach to an invalid path would fail. Do you mean to write a test that binds to each endpoint just to prove that each exists? @jwolski

lib/trace/core.js
benfleis: Why would we bother here? Sub/unsub is infrequent and ought to have no impact on performance. They could be preloaded, but it also seems a waste of time. @jwolski

lib/trace/core.js
benfleis: This was unclear. The class-level is now renamed to validateOptions(), since the intent to pre-check opts during subscription handling. The instance-levels check with instances.

lib/trace/core.js
benfleis: The config is indeed exported, but I disagree on this point -- resolution is an action, not passive config. As previously mentioned it could be loaded during init, but not at compile/load time, so I argue that it's a core func, even if not repeatedly called. Do you buy it? @jwolski

lib/trace/index.js
benfleis: agreed, done.

lib/trace/index.js
benfleis: gone, esp. since the registration is now moved.

lib/trace/index.js
benfleis: done.

lib/trace/index.js
benfleis: fix0red.

lib/trace/index.js
benfleis: agreed. done.

lib/trace/index.js
benfleis: fix0red.

lib/trace/index.js
benfleis: gone due to other reasons, but agreed.

lib/trace/index.js
benfleis: fix0red.

lib/trace/index.js
benfleis: fix0red.

lib/trace/index.js
benfleis: fix0red.

lib/trace/index.js
benfleis: done.

lib/trace/index.js
benfleis: yes. done.

lib/trace/index.js
benfleis: we talked about it, and your edit is mostly correct. Tracer brings together a node internal event emitter/listener, and forwards those to the sink. I think we found no better naming, but I have added a small comment to this effect at the top of core/Tracer.

lib/trace/index.js
benfleis: done.

lib/trace/index.js
benfleis: gone, and I reworked the types code to have objects hang from the map, and skip that already unneeded default case. it was a bit of leftover from a previous thing, and needed ousting.

lib/trace/index.js
benfleis: I changed it to traceEvent (as we discussed), if that still sounds disagreeable, let's chat. @jwolski

lib/trace/index.js
benfleis: in v0.0.1b! this is v0.0.1a ;)

lib/trace/index.js
benfleis: They serve totally different purposes, and since they're only called in 1 place, I saw no reason to make a parent to do both. I'll add one if/when that changes.

lib/trace/index.js
benfleis: .cz

lib/trace/index.js
benfleis: good catch.

lib/trace/index.js
benfleis: done

lib/trace/index.js
benfleis: same response as above. could be, no need, yet.

lib/trace/index.js
benfleis: done

lib/trace/log.js
benfleis: Buffer is solved by that, but what about sending plain ol' objects. I am inclined to support it directly, or to add an 'asJSON' option to the registration. Whatcha think? @jwolski

lib/trace/tchannel.js
benfleis: did _remoteCallback. it's not great, but at least neutral in name, to match its intent. @jwolski

lib/trace/tchannel.js
benfleis: That is the current assumption. I don't tend to agree with the idea of splitting up -- if you want lots of data in that 1 arg, wrap it into an object with labels. I'm definitely open to discussion on this one; I don't have a strong opinion. @jwolski

lib/trace/tchannel.js
benfleis: see above. json encoding vs. toString?

lib/trace/tchannel.js
benfleis: people debugging. i err on the side of extra events, cuz they're super handy at times.

lib/trace/tchannel.js
benfleis: (discussed in person)

I argue for idempotence over the full src-dst pairing set { event, sinkType, sinkEndpoint, sinkURL, etc. }, since anything else allows duplicate sending to the same thing. Which is probably bad. And relying on a token means that one cli tool can't easily clean up after another.

index.js
```diff
@@ -153,6 +154,8 @@ function RingPop(options) {
 
     createEventForwarder(this);
 
+    this.tracers = new TracerStore();
```
jwolski: legit

server/trace.js
```diff
@@ -0,0 +1,66 @@
...
+};
+
+function createAddTrace(ringpop) {
+    return function addTrace(optsStr, _, hostInfo, cb) {
+        _updateTracers('add', optsStr, cb);
+    }
+}
+
+function createRemoveTrace(ringpop) {
+    return function addTrace(optsStr, _, hostInfo, cb) {
```
jwolski: Copy/paste errors?? This should be removeTrace.

server/trace.js
```diff
@@ -0,0 +1,66 @@
...
+        _updateTracers('add', optsStr, cb);
+    }
+}
+
+function _updateTracers(ringpop, fnStr, optsStr, cb) {
+    var opts = toCamelCase(safeParse(optsStr));
+    var config = traceCore.resolveEventConfig(ringpop, opts.event);
+    if (!config) {
+        var err = errors.InvalidOptionError({ option: 'event', reason: 'it is invalid'});
+        ringpop.logger.warn('invalid event', {
```
jwolski: Can you preface this with logger message with "ringpop". Also "invalid event" is very vague. Is it possible to provide a bit more context here?

test/trace-test.js
```diff
@@ -0,0 +1,135 @@
...
+        assert.equal(ringpop.journal[0], 'foo', 'journal added foo');
+
+        ringpop.membership.emit('updated', 'bar');
+        assert.equal(ringpop.journal.length, 2, 'journal has 2 entries');
+        assert.equal(ringpop.journal[1], 'bar', 'journal added bar');
+    });
+});
+
+test('trace events can renew, and expire', function t(assert) {
+    var ringpop = new rpMock();
```
jwolski: Can you make this look like a typical constructor (aka RpMock or RingpopMock)

test/trace-test.js
```diff
@@ -0,0 +1,135 @@
...
+    assert.plan(3);
+
+    // nextTick okay since log "connects" instantly
+    process.nextTick(function onTracer() {
+        // now test events on this thing
+        ringpop.membership.emit('updated', 'foo');
+        assert.equal(ringpop.journal.length, 1, 'journal has 1 entry');
+
+        // renew after 25, run @ 60, confirm event; this is racey...
+        setTimeout(store.add.bind(store, config, opts), 25);
```
jwolski: Can you use `time-mock` to advance time instead of slowing down tests

test/trace-test.js
```diff
@@ -0,0 +1,135 @@
...
+            assert.equal(ringpop.journal.length, 2, 'journal unchanged');
+        }, 100);
+    });
+});
+
+// dumbest of mocks for our trace element
+function rpMock() {
+    var journal = [];
+    this.journal = journal;
+    this.logger = {
```
jwolski: Using logger calls to validate assertions seems error prone. While I understand that you're using the log tracer which writes to the log, any logging statements that are introduced will throw these tests off. Can you instead attach a listener to the send event of the LogTracer and journal that way?

lib/trace/store.js
```diff
@@ -0,0 +1,143 @@
...
+    var timer = tracer._timer;
+    var remove = this.remove.bind(this, config, opts);
+
+    // 0/null offset implies already expired, thus don't re-add
+    if (timer) {
+        clearTimeout(tracer._timer);
+        tracer._timer = null;
+    }
+    if (offset) {
+        tracer._timer = setTimeout(remove, offset);
```
jwolski: TracerStore should offer up a `destroy()` function to clear all existing tracer timers. TracerStore can listen to the `destroyed` event that Ringpop emits or call tracerStore.destroy() from Ringpop's destroy() function. The former is the convention I've been using lately.

server/trace.js
```diff
@@ -0,0 +1,66 @@
...
+    },
+    removeTrace: {
+        endpoint: '/trace/remove',
+        handler: createRemoveTrace,
+    },
+};
+
+function createAddTrace(ringpop) {
+    return function addTrace(optsStr, _, hostInfo, cb) {
+        _updateTracers('add', optsStr, cb);
```
jwolski: you're going to want to use tchannel's arg3 as the body and not arg2. I believe these handlers should read:

```
return function addTrace(arg2, arg3, hostInfo, cb) {
  _updateTracers('add', arg3, cb);
}; 
```

server/trace.js
```diff
@@ -0,0 +1,66 @@
...
+        handler: createAddTrace,
+    },
+    removeTrace: {
+        endpoint: '/trace/remove',
+        handler: createRemoveTrace,
+    },
+};
+
+function createAddTrace(ringpop) {
+    return function addTrace(optsStr, _, hostInfo, cb) {
```
jwolski: it's strange to `_` one are and not the other (hostInfo) that you're not using. I think it also gets weird to start calling them `_` and `_2`. I'd prefer that you call them by actual names.

lib/trace/store.js
```diff
@@ -0,0 +1,143 @@
...
+
+        tracer.connectSink(function onConnect(err) {
+            if (err) {
+                self.remove(config, opts, null);
+                self.emit('add', err, null);
+                return;
+            }
+            tracer.connectSource();
+            self._updateExpiration(config, opts, tracer);
+            self.emit('add', null, tracer);
```
jwolski: I think emitting `add` like this makes for a strange API. Either you emit synchronously or async, but not both depending upon the situation. The sync emit() below deserves a nextTick. Thoughts?

jwolski:
I left a few more comments for you in my 2nd pass review. Let me know when 2nd pass feedback has been addressed.

test/trace-test.js
benfleis: Gladly. I hadn't seen time-mock (I looked briefly at stealing the time mock from sinon), and ultimately thought that these tiny timevals wouldn't be so bad, but I can easily rework to use the mock.

(Just checked, they add 140ms total time to the tests as-is. I still feel like mock time is better here for reliable timing, instead of the risk of system delays screwing up tests. But just thought I'd share.)

lib/trace/store.js
benfleis: Totally agreed, I missed this one. (I also try to do all sync or all async.)

server/trace.js
benfleis: Hrm. Leftovers. I agree w/ _ and _2 -- I miss languages that allow the duplicate _, _, convention. I will give them names, prefixed w/ _ (which I know as another convention for indicating ignored args; can also do _ign1 and _ign2, which is less weird than _ and _2).

test/trace-test.js
benfleis: I don't understand how the logging statements would throw something off, unless you mean by their absence.

RingpopMock is required to have a logger of some sort; if I don't override, the debugging lines end up as real log outputs. I could filter based on the actual input, and allow non-special lines through, but this seemed the simplest solution to me. (If a bug comes up, the dev might have to consider another logger mechanism. I also thought about adding an optional 'logger' object to the LogTracer object itself, allowing a totally independent logger object.)

Preference, @jwolski?

test/trace-test.js
benfleis: yep, sorry.

server/trace.js
benfleis: fixed

server/trace.js
benfleis: the primal ape in me wants to smash something with "arg2" and "arg3". are they supposed to be header and body? it's a strange and unpredictable convention (to the uninitiated). @jwolski 

and yes, i'll fix :)

server/trace.js
benfleis: yes, now "unknown trace event requested". it's not great, but definitely better.

lib/trace/store.js
jwolski: Bumpity

server/trace.js
jwolski: yes, it's also conventional to call them `head` and `body`

test/trace-test.js
jwolski: I think since this test is executing within a very controlled (stubbed/mocked) environment, the assertion on the logger calls are fine.

lib/trace/store.js
benfleis: Heh, this is also done, I just forgot to write about it here :)

benfleis:
@wolski the logs are fixed, I actually want to pull that event generator code out for tidiness. I will do a micro benchmark for kicks :)

benfleis:
@jwolski oh yeah, I should also note that I am putting off the conflict resolve until we're through this whole thing!

PR #142, opened 2015-08-17
Implement tchannel forwarding for node.

Raynos:
This is an implementation of ringpop forwarding for
tchannel v2.

You can see the expected use case in the examples.

TODO: 

 - [ ] ringpop forwarding stats
 - [ ] early abort on checksum failure
 - [ ] retries?

I suspect we do not want retries in RingpopHandler and
should rely on retries in either the client or in hyperbahn
itself. We currently have retries in HTTP forwarding because
we dont have retries in HTTP clients..

Adding early abort on checksum failure is possible but I should
probably carve out another transport header for it. Maybe
add "server checksum" or "sc" as a transport header.

If we do not implement early abort on checksum failure then
a request could in theory get forwarded forever because
infinite divergence or flappy-ness. Of course it will timeout
in tchannel itself but it would be nicer to have a UnexpectedError
telling you that its ringpop divergence instead.

This code is already used in two applications at uber so it's
time we add it to ringpop core itself.

r: @jwolski @jcorbin @benfleis
cc: @rssathe @rf

### DIFF
--- a/ringpop-handler.js
+++ b/ringpop-handler.js
@@ -84,7 +84,8 @@ RingpopHandler.prototype.handleRequest = function handleRequest(req, buildRes) {
             serviceName: req.serviceName,
             endpoint: req.endpoint,
             callerName: req.headers.cn,
-            headers: req.headers
+            headers: req.headers,
+            local: self.ringpop.whoami()
         });
         return buildRes().sendError(
             'BadRequest',
--- a/ringpop-handler.js
+++ b/ringpop-handler.js
@@ -1,0 +1,108 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var assert = require('assert');
+
+var RelayRequest = require('tchannel/relay_handler').RelayRequest;
+
+module.exports = RingpopHandler;
+
+function RingpopHandler(options) {
+    /*eslint max-statements: [2, 25]*/
+    if (!(this instanceof RingpopHandler)) {
+        return new RingpopHandler(options);
+    }
+
+    var self = this;
+
+    self.realHandler = options.realHandler;
+    assert(
+        typeof self.realHandler === 'object',
+        'expected options.realHandler'
+    );
+
+    self.ringpop = options.ringpop;
+    assert(typeof self.ringpop === 'object', 'expected options.ringpop');
+
+    self.channel = options.channel;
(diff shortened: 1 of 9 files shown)


rssathe:
@Raynos I was confused as to why the Ringpop handler took a blacklist instead of a whitelist. Also ideally when you register an endpoint you can pass some flag saying you want this endpoint to be blacklisted/whitelisted instead of having duplicating logic when registering arg1s and also passing in a list of arg1s to the ringpop handler

rssathe:
For the blacklist vs whitelist thing can we safely assume that most endpoints should have a shard key?

Raynos:
@rssathe 

What we found in production is that the majority of endpoints are sharded except for `health`.

You can also remember that tchannel supports multiple service names so you could add a ringpop handler for `my-app` serviceName and have a `my-app-admin` serviceName that does not use a ringpop handler and is not sharded.

test/lib/alloc-ringpop.js
```diff
@@ -64,8 +64,8 @@ function allocRingpop(name, options) {
 }
 
 function semiRandPort() {
-    var base = 20000;
-    var rand = Math.floor(Math.random() * 20000);
+    var base = 10000;
```
jwolski: wat

ringpop-handler.js
```diff
@@ -0,0 +1,108 @@
...
+    var self = this;
+
+    if (self.blacklist && self.blacklist[req.endpoint]) {
+        return self.realHandler.handleRequest(req, buildRes);
+    }
+
+    var shardKey = req.headers.sk;
+    if (!shardKey) {
+        self.logger.warn('Ringpop got request without a shardKey', {
+            socketRemoteAddr: req.remoteAddr,
```
jwolski: add `local: self.ringpop.whoami()` to this logging meta

ringpop-handler.js
```diff
@@ -0,0 +1,108 @@
...
+    var dest = self.ringpop.lookup(shardKey);
+    if (self.ringpop.whoami() === dest) {
+        return self.realHandler.handleRequest(req, buildRes);
+    }
+
+    var outreq = new RelayRequest(self.channel, req, buildRes);
+    outreq.createOutRequest(dest);
+};
+
+RingpopHandler.prototype.register = function register(arg1, fn) {
```
jwolski: why is this function definition requred?

test/integration/proxy-test.js
```diff
@@ -978,7 +978,8 @@ test('handle tchannel failures', function t(assert) {
         setTimeout(function onTimer() {
             var name = cluster.two.channel.hostPort;
             var peer = cluster.one.channel.peers.get(name);
-            var conn = peer.connections[0]; // the "out" one
+
```
jwolski: *shrug*

ringpop-handler.js
```diff
@@ -0,0 +1,108 @@
...
+            '[ringpop] Request does not have sk header set'
+        );
+    }
+
+    var dest = self.ringpop.lookup(shardKey);
+    if (self.ringpop.whoami() === dest) {
+        return self.realHandler.handleRequest(req, buildRes);
+    }
+
+    var outreq = new RelayRequest(self.channel, req, buildRes);
```
jwolski: It may be the case, if the ring is in a very inconsistent state, that forwarded requests bounce around indefinitely. Is there a TTL handled internal to TChannel?

ringpop-handler.js
```diff
@@ -0,0 +1,108 @@
...
+RingpopHandler.prototype.type = 'tchannel.endpoint-handler';
+
+RingpopHandler.prototype.handleRequest = function handleRequest(req, buildRes) {
+    var self = this;
+
+    if (self.blacklist && self.blacklist[req.endpoint]) {
+        return self.realHandler.handleRequest(req, buildRes);
+    }
+
+    var shardKey = req.headers.sk;
```
jwolski: any intention of supporting multiple shard keys in the req headers? Some services require it.

jwolski:
I left a few comments/questions. Thanks for the contribution.

ringpop-handler.js
jwolski: NVM: I have seen the statement in the description of this PR.

ringpop-handler.js
Raynos: `sk` is just a string; we can structure it however we want.

ringpop-handler.js
Raynos: tchannel has logic where `tchannel#register` calls `self.handler.register()`

test/integration/proxy-test.js
Raynos: This is part of the "listen on top channel only" fix necessary; see seperate commit.

test/lib/alloc-ringpop.js
Raynos: Tests were flaky on linux by picking a random port in the ephemeral range.

By picking a random port in the non-ephemeral range the tests no longer fail with EADDRINUSE

Raynos:
I think to support more services we might need a application specific `resolveHost`

test/lib/alloc-ringpop.js
jwolski: sgtm

PR #144, opened 2015-08-21
Remove readable-stream dep

jwolski:
I can't imagine this is helping Ringpop any either: https://github.com/uber/tchannel/pull/1179. 

@Raynos @jcorbin


Raynos:
lgtm.

PR #145, opened 2015-08-24
Move admin handler to their own dir

jwolski:
This might look like a lot of shuffle. But there's 0 functionality change (famous last words). Though completely unrelated, it's in preparation for landing damp scoring. The reason this change exists is because a new admin handler is needed in flap damping to apply arbitrary membership updates to a node. That handler is still to come (in another PR), but I think this re-org makes the admin API more intuitive.

In short, this places all admin API endpoint handlers into a `server/admin` directory. And the `index.js` within defines both the URL/endpoint and the handler function for each.

@danielheller @CorgiMan @benfleis @Raynos


package.json
```diff
@@ -52,7 +52,7 @@
     "pre-commit": "^0.0.9",
     "tape": "^3.0.3",
     "tape-cluster": "2.1.0",
-    "tchannel": "^2.5.1",
+    "tchannel": "2.5.1",
```
jwolski: This is a dev dependency of tchannel. A recent tchannel update broke tests. This is in an effort to pass this test run and potentially fix in another PR. You can ignore this change for now.

server/admin/debug-clear.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
jwolski: NOTE FOR REVIEWER: same functionality as before. just moved out of `server/index.js` and into its own file.

server/admin/debug-set.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
```
jwolski: same functionality as before. just moved out of `server/index.js` and into its own file.

server/admin/join.js
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
```
jwolski: same functionality as before. just moved out of `server/index.js` and into its own file.

server/admin/leave.js
```diff
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 'use strict';
 
-var errors = require('../lib/errors.js');
+var errors = require('../../lib/errors.js');
```
jwolski: same functionality as before. just moved out of `server/index.js` and into its own file.

server/admin/lookup.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
```
jwolski: same functionality as before. just moved out of `server/index.js` and into its own file.

server/admin/reload.js
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
```
jwolski: same functionality as before. just moved out of `server/index.js` and into its own file.

server/admin/stats.js
```diff
@@ -19,8 +19,8 @@
 // THE SOFTWARE.
 'use strict';
 
-module.exports = function handleAdminLookup(opts, callback) {
-    callback(null, null, {
-        dest: opts.ringpop.lookup(opts.key)
-    });
+module.exports = function createHandler(ringpop) {
```
jwolski: same functionality as before. just renamed file

server/admin/tick.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
```
jwolski: same functionality as before. just moved out of `server/index.js` and into its own file.

benfleis:
@jwolski 

Overall questions:
 - same functionality ?= same code, just shifted (It could mean reimplemented w/ same in/out, in which case I should be reading the actual content as well.)
 - all of the createHandle and handle lambda fns lost their name specificity. the filename will show in the trace, but it might still be nice to have the redundant fn names in their. Thoughts?

Assume above `functionality === code`, then lgtm. If not, then I'll give a better read pass.

jwolski:
good point re: fn names. I will fix.

jwolski:
sorry. and yes, it is the "same code".

jwolski:
actually, there was a slight change in endpoint registration logic (to make it generic for both admin + protocol endpoints). you can take a look at that here: https://github.com/uber/ringpop/pull/145/files#diff-0861d6d6b50d7d695344bf2d86d6e5e6R42

PR #146, opened 2015-08-24
Remove MemberUpdate class

jwolski:
I'm not sure why I thought this was a great idea. Only used in a test file. Removing it for now. It only makes things more confusing. This is another PR in prep for flap damp scoring.

@CorgiMan @benfleis @danielheller @Raynos


jwolski:
Self-merging. If you are grossly offended, leave a comment.

PR #148, opened 2015-08-24
Beef up the Member class

jwolski:
This is also in preparation for landing flap damp scoring. The scoring goes on in the member.js file. And as such, it makes sense to bake a lot of the other membership related details (like evaluating override rules into it too).

@CorgiMan @danielheller @benfleis @Raynos


jwolski:
thanks @benfleis

jwolski:
arg. I lost your comments since squashing. Please see your email history for how they've been addressed. I'll not do that next time :dash:

jwolski:
Re-opening against master.

PR #152, opened 2015-08-25
Close all peer connections to trigger socket closed response

jwolski:
This is a fix for the test that broke in between tchannel (as a devDependency) between 2.5.1 and 2.6.1. I confirmed with @jcorbin that the tchannel commit that git-bisect pointed at was intentional. Thankfully, this "problematic" commit neither breaks Ringpop nor TChannel, just a test (that was highly coupled to existing TChannel behavior). This closes all peer connections and induces a closed socket and makes the test pass.

@benfleis @danielheller @CorgiMan @Raynos


jcorbin:
lgtm

PR #153, opened 2015-08-26
Run integration tests

jwolski:
This is a convenience that I needed after debugging a timer leak in our test suite. It makes sense to separate out this change from others.


benfleis:
i'm all for these sorts of convenience separation. lgtm.

PR #158, opened 2015-08-27
Flap damp scoring

jwolski:
This is the first of several PRs to come for flap damping. The first is damp scoring. A member in the membership list will be penalized for every update that is applied. When the damp score exceeds the configured damp suppress limit, an event will be emitted. Later, this event will initiate the damping subprotocol. (That's the next PR to come). The damp scores for all members decay exponentially, its periodicity governed by another configurable parameter. 

I left a few cheeky comments and added a few bells/whistles (like the separate Config object) to this PR to accompany the scoring. Don't mind them too much.

The test coverage is fairly reasonable.

Oh, and also... tick-cluster needed a bit of fixing. A member's damp score is now part of the returned stats. The way tick-cluster previously computed checksums for membership was too restrictive.

There's actually no net gain in terms of functionality in this PR. It's simply a mechanism to start scoring members and learning about scoring, in general.

@danielheller @benfleis @CorgiMan

### DIFF
--- a/config.js
+++ b/config.js
@@ -40,8 +40,8 @@ Config.prototype.get = function get(key) {
 Config.prototype.set = function set(key, value) {
     var oldValue = this.store[key];
     this.store[key] = value;
-    this.emit('changed', key, value, oldValue);
-    this.emit('changed.' + key, value, oldValue);
+    this.emit('set', key, value, oldValue);
+    this.emit('set.' + key, value, oldValue);
 };
 
 Config.prototype._seed = function _seed(seed) {
--- a/config.js
+++ b/config.js
@@ -1,0 +1,75 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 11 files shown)


config.js
```diff
@@ -0,0 +1,75 @@
...
+util.inherits(Config, EventEmitter);
+
+Config.prototype.get = function get(key) {
+    return this.store[key];
+};
+
+Config.prototype.set = function set(key, value) {
+    var oldValue = this.store[key];
+    this.store[key] = value;
+    this.emit('changed', key, value, oldValue);
```
benfleis: perhaps 'update' or 'set', since those are the action labels we use elsewhere?

lib/membership/member.js
```diff
@@ -28,10 +28,39 @@ function Member(ringpop, update) {
     this.address = update.address;
     this.status = update.status;
     this.incarnationNumber = update.incarnationNumber;
+    this.dampScore = update.dampScore || ringpop.config.get('dampScoringInitial');
```
benfleis: can update.dampScore === 0?

lib/membership/member.js
```diff
@@ -59,8 +88,24 @@ Member.prototype.evaluateUpdate = function evaluateUpdate(update) {
...
+        // will be penalized. The scoring levers will control persistent
+        // flaps. We'll eventually get _real_ good at identifying flaps
+        // and apply penalties more strictly.
+        this._applyUpdatePenalty();
+        this.lastUpdateDampScore = this.dampScore;
+    }
+
     this.emit('updated', update);
 
+    // lastUpdateTimestamp must be updated after the penalty is applied
```
benfleis: I <3 this kind of comment.

test/unit/config_test.js
```diff
@@ -0,0 +1,80 @@
...
+    assert.plan(2);
+
+    var config = new Config();
+    config.on('changed', onConfigChanged);
+    var configKey = 'somekey';
+    config.on('changed.' + configKey, onConfigKeyChanged);
+    var configVal = 'someval';
+    config.set(configKey, configVal); // this triggers change events *fingers crossed*
+
+    assert.end();
```
benfleis: Maybe I misunderstand .pass(), but I don't think this test can fail? would it be better to set count = 0, and have each event callback incr, and assert count == 2?

benfleis: Ahh, I see. .plan tells it to require 2 asserts. Ignore me, since .plan + .pass does exactly what I described :)

benfleis:
done with this round. only potential bug is the explicit == 0 possibility above. other than that and 'changed', lgtm

config.js
jwolski: yea, i think that's a good suggestion. ill "change" it ;)

lib/membership/member.js
jwolski: yes, good catch. Since dampScores are not yet disseminated, this isn't an issue...yet. I'll fix no matter.

PR #159, opened 2015-08-28
add a skipLookupOnRetry request opt

weikai77:
@jwolski @markyen This allows ringpop-replicator to disable key lookups on retries, and hence fixes the following issue:

> Before retrying, ringpop looks up the destination for the given keys again, and if it finds the keys are
> now mapped to more than one destination, ringpop will return the "Destinations for proxied request 
> have diverged" error. This behavior is not compatible with replication because the lookup during 
> retries does not account for the replicas destinations.


jwolski:
lgtm

PR #160, opened 2015-09-02
Add make-* stats

jwolski:
Ringpop does not currently emit stats when a node explicitly changes the status of a member. This new stat is complementary to the `membership-update.*` stats that Ringpop emits when it observes an update that has been gossiped and applied.

@benfleis @danielheller @Raynos @CorgiMan


benfleis:
Other than my pedantry, lgtm.

jwolski:
Your point is well-taken. "mark" might have been a slightly better choice. One could argue that an 'update is being made.' I'm going to push this one through as is. Come hunt me down if you have objections.

PR #162, opened 2015-09-11
Join blacklist to prevent unintended hosts from joining cluster

jwolski:
This is the start of a formal member blacklist. This prevents joins from 'unintended' members which are specifiable by a `memberBlacklist` config. Apps can choose arbitrary regexs for the blacklist.

Oh, and yes, this'll need to be back-ported, if services that desire this functionality cannot upgrade to 10.x.x.

@CorgiMan @danielheller @markyen @benfleis

### DIFF
--- a/index.js
+++ b/index.js
@@ -121,7 +121,7 @@ function RingPop(options) {
 
     // Initialize Config before all other gossip, membership, forwarding,
     // and hash ring dependencies.
-    this.config = new Config(options);
+    this.config = new Config(this, options);
 
     this.requestProxy = new RequestProxy({
         ringpop: this,
@@ -249,8 +249,8 @@ RingPop.prototype.bootstrap = function bootstrap(opts, callback) {
         return;
     }
 
-    this.checkForMissingBootstrapHost();
-    this.checkForHostnameIpMismatch();
+    checkForMissingBootstrapHost();
+    checkForHostnameIpMismatch();
 
     // Add local member to membership.
     this.membership.makeAlive(this.whoami(), Date.now());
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -249,8 +249,8 @@ RingPop.prototype.bootstrap = function bootstrap(opts, callback) {
         return;
     }
 
-    this.checkForMissingBootstrapHost();
-    this.checkForHostnameIpMismatch();
+    checkForMissingBootstrapHost();
```
jwolski: This is incidental cleanup. I very much disliked how these were part of the prototype.

lib/swim/join-sender.js
```diff
@@ -113,11 +112,6 @@ function JoinCluster(opts) {
     this.maxJoinDuration = numOrDefault(opts.maxJoinDuration,
         MAX_JOIN_DURATION);
 
-    // We want to abort the joining process if we have failed
```
jwolski: I moved this param into `config.js` so that I can make the integration tests "faster" and also because all configs should eventually belong in `config.js`

server/protocol/join.js
```diff
@@ -59,7 +68,22 @@ function validateJoinerAddress(ringpop, joiner, callback) {
...
+            joiner: joiner,
+            blacklist: blacklist
+        }));
+        return false;
+    }
+
     return true;
+
+    function anyBlacklisted() {
+        return _.any(blacklist, function any(pattern) {
```
markyen: Since you already verified this is an Array, why not just use `Array.prototype.some`? If you add an `isArray` check to the validator, you could also use `Array.prototype.every` there

jwolski: TIL `some()` existed.

server/protocol/join.js
```diff
@@ -19,9 +19,18 @@
 // THE SOFTWARE.
 'use strict';
 
+var _ = require('underscore');
```
markyen: Unused import

PR #163, opened 2015-09-11
/admin/config/set and /admin/config/get endpoints

jwolski:
Admin endpoints that allow you to change config dynamically. Implementation-wise this is pretty done. I'll add a couple of integration tests.

This functionality is not immediately needed, but was a fun Friday evening project before I went home. And this'll give us a reason to add `ringpop-admin config` commands. Have a nice weekend! 

### DIFF
--- a/client.js
+++ b/client.js
@@ -22,13 +22,8 @@
 var safeParse = require('./lib/util.js').safeParse;
 var TChannel = require('tchannel');
 
-function RingpopClient(host, port) {
-    this.host = host || '127.0.0.1';
-    this.port = port || 2999;
-    this.tchannel = new TChannel({
-        host: this.host,
-        port: this.port
-    });
+function RingpopClient() {
+    this.tchannel = new TChannel();
     this.subChannel = this.tchannel.makeSubChannel({
         serviceName: 'ringpop'
     });
--- a/client.js
+++ b/client.js
@@ -1,0 +1,82 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+'use strict';
+
+var safeParse = require('./lib/util.js').safeParse;
+var TChannel = require('tchannel');
+
+function RingpopClient(host, port) {
+    this.host = host || '127.0.0.1';
+    this.port = port || 2999;
+    this.tchannel = new TChannel({
+        host: this.host,
+        port: this.port
+    });
+    this.subChannel = this.tchannel.makeSubChannel({
+        serviceName: 'ringpop'
+    });
+}
+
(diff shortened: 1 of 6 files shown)


benfleis:
I have failed you. I'll have a look at this despite already being
committed. And the other one.

b

jwolski:
:) np. have a look at your convenience. ill address any concerns you have.

test/integration/admin_test.js
```diff
@@ -0,0 +1,53 @@
...
+}, 'config endpoints', function t(bootRes, cluster, assert) {
+    assert.plan(4);
+
+    var client = new RingpopClient();
+    async.series([
+        function configSetPart(callback) {
+            client.adminConfigSet(cluster[0].whoami(), {
+                testconfig1: 1
+            }, function onSet(err) {
+                assert.notok(err, 'no error occurred');
```
benfleis: style nit for here and below -- when checking for no error, I find .notok(err) to break my tiny brain in 2. tape doesn't offer a 'noterror' or 'noerror' predicate, so I would suggest .false, which is at least gives me a direct notion of what to expect.

test/unit/server/admin/config_test.js
```diff
@@ -0,0 +1,84 @@
...
+
+test('config set handler validation', function t(assert) {
+    var ringpop = new Ringpop({
+        app: 'ringpop',
+        hostPort: '127.0.0.1:3000'
+    });
+    var handleConfigSet = createConfigSetHandler(ringpop);
+    // Missing body
+    handleConfigSet(null, JSON.stringify(null), null, function onHandle(err) {
+        assert.ok(err, 'an error occurred');
```
benfleis: like the other comment, `ok(err)` also hurts my head. ok to me implies no error, but we're asserting the presence of an error, and thus, I suggest `true(err)`.

client.js
```diff
@@ -0,0 +1,77 @@
...
+    }, function onIdentified(err) {
+        if (err) {
+            callback(err);
+            return;
+        }
+
+        self.subChannel.request({
+            host: host,
+            hasNoParent: true,
+            retryLimit: 1,
```
benfleis: This feels low. Is there a reason to turn this down from the more typical 3 that I've seen elsewhere?

server/admin/config.js
```diff
@@ -0,0 +1,74 @@
...
+function createConfigGetHandler(ringpop) {
+    return function handleConfigGet(arg2, arg3, hostInfo, callback) {
+        var body = safeParse(arg3 && arg3.toString());
+
+        if (!body || !Array.isArray(body)) {
+            callback(null, null, JSON.stringify(ringpop.config.getAll()));
+            return;
+        }
+
+        var configs = body.reduce(function each(memo, key) {
```
benfleis: perhaps in the future we could add the `getIn` functionality that I've got in the still unmerged event PR :)

server/admin/config.js
```diff
@@ -0,0 +1,74 @@
...
+            memo[key] = ringpop.config.get(key);
+            return memo;
+        }, {});
+
+        callback(null, null, JSON.stringify(configs));
+    };
+}
+
+function createConfigSetHandler(ringpop) {
```
benfleis: This is not `set`, but `merge` (or extend, update, etc.). In the vein of self documenting code/API, this is dangerous.

benfleis:
Done with review. @jwolski

test/integration/admin_test.js
jwolski: i hate notok too. Honestly, I'm less in love with `false`. ok/notok at least calls out to me that it's checking whether the value 'exists'. `false()` just says to be we're checking boolean true or false (even though I know it's really truthiness/falsiness). I'm not in love with the tape API in general.

test/unit/server/admin/config_test.js
jwolski: ack

client.js
jwolski: I'm not sure what the recommendation from TChannel is. 1 feels lows. But this client library is also not used for anything other than tests. I kept it client.js general to eventually support alternative uses (like outside of the test environment).

server/admin/config.js
jwolski: yes, we can make a util out of it. Maybe underscore provides this utility already.

server/admin/config.js
jwolski: I think get/set are much more familiar idioms. And am using these sort of APIs as inspiration:

http://redis.io/commands/config-set
https://docs.npmjs.com/cli/config#set

client.js
benfleis: It lives in root, so I can't help but believe it will get used, as would that request. If it's test intended, then I vote for moving it there.

server/admin/config.js
benfleis: I don't want to paint the barn! But I will finish my thought and be done.

I didn't realize it at the time, but do now. I have qualms w/ those APIs for the same reason: there is an implicit "-key" at the end of those. I set "set config" as meaning set an entire config, and "set config key" as setting a single key within a config, optionally taking an array of keys/values as a batch op.

/me out.

client.js
jwolski: It's general purpose. Neither coupled to tests nor core. Root is an OK place. It could also live in lib. But the rules around lib and root are murky.

server/admin/config.js
jwolski: ack.

But I disagree. I think this implicit key is a firmly established, well-known model.

PR #164, opened 2015-09-13
/admin/gossip/start, stop, tick endpoints

jwolski:
Just another quick enhancement to admin API for full control of gossip (start, stop, tick).

@CorgiMan @benfleis @danielheller


test/unit/server/admin/gossip_test.js
```diff
@@ -19,10 +19,29 @@
...
+var createGossipStartHandler = gossipHandlers.gossipStart.handler;
+var createGossipStopHandler= gossipHandlers.gossipStop.handler;
+
+test('start/stop gossip', function t(assert) {
+    var ringpop = new Ringpop({
+        app: 'ringpop',
+        hostPort: '127.0.0.1:3000',
+        autoGossip: false
+    });
+    var handleGossipStart = createGossipStartHandler(ringpop);
```
benfleis: maybe throw in a small test first to confirm that gossip is indeed off? might be overkill.

benfleis:
lgtm

test/unit/server/admin/gossip_test.js
jwolski: Yea, I wasn't sure how far to take it.

I've got start/stop tests that assert on the `isStopped` flag. This seemed 'good enough' at the time. These are unit tests so I was leaving the test to make sure that 'gossip is indeed stopped' for the Gossip unit tests.

PR #165, opened 2015-09-16
Allow body limit to be specified for large object payloads

jwolski:
Cherry-pick 'the body module' fix applied to Ringpop v9 to master.

@jwyuan


charliezhang:
lgtm. thanks

PR #166, opened 2015-09-17
example: using Hyperbahn discovery API for bootstrap

ShanniLi:
r @benfleis 
cc @jwolski @Raynos 

There is one issue:
If you run this example more than once without restarting Hyperbahn, Ringpop will refuse to join, since the host list returned by discovery API has dead hostports from previous runs.

### DIFF
--- a/examples/discovery.js
+++ b/examples/discovery.js
@@ -1,0 +1,217 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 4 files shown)


Raynos:
This raises an important "feature" of Hyperbahn::discover

You will discover all peers; dead and alive. Hyperbahn does not return any information about whether a peer is alive.

examples/discovery.js
```diff
@@ -0,0 +1,217 @@
...
+            cb([]);
+            return;
+        }
+
+        var hosts = [];
+        for (var i = 0; i < res.body.peers.length; i++) {
+            hosts.push(covertHost(res.body.peers[i]));
+        }
+
+        cb(hosts);
```
Raynos: callbacks have two arguments.

examples/discovery.js
```diff
@@ -0,0 +1,217 @@
...
+    }
+};
+
+App.prototype.bootstrap = function bootstrap(cb) {
+    var self = this;
+
+    // listen
+    self.channel.listen(0, '127.0.0.1', onListen);
```
jwolski: Listening on `0` is not going to jive well with Ringpop, as it is currently implemented today. On restart, the node that previously identified itself as host:port `X:Y` needs to be able to preserve that identity. This is not guaranteed in the port `0` model as far as I understand unless Ringpop builds in a higher-level identification mechanism that is different than its listening address. I think that's the first problem to solve, then the port 0 thing will be not a problem.

Issue #167, opened 2015-09-21
checksum not recalculated when 2-node ring loses 1 member


benfleis:
Perhaps this is a corner case; I need to read the code, but wanted to note that I've seen this multiple times, and find it counterintuitive.

PR #169, opened 2015-09-21
allow 2-node clusters to split and rejoin

benfleis:
addresses: https://github.com/uber/ringpop-node/issues/167

Allow a 2-node cluster to split and generate proper suspect mark events, etc.,
and subsequent rejoins. Current code short circuits suspect logic, and will
never mark faulty or generate new checksum.

Since there is a test demonstrating that 2-node ping fail generates an error,
not sure if this is the correct/desired approach.


jwolski:
love it! lgtm

PR #171, opened 2015-09-25
emit destroying message right before ringpop is destroyed

ShanniLi:
r @jwolski @benfleis 

This is necessary for whoever needs to use channel to clean up before it is destroyed.


jwolski:
lgtm

PR #182, opened 2015-09-30
Get rid of the dreadful pingMemberNow; make tick() part of gossip

jwolski:
This pingMemberNow has been bothering me since the dawn of time. I had to take the plunge and get rid of it. Minor refactoring. Just moved some code around, leading to a more intuitive Gossip interface.

@uber/ringpop


benfleis:
yay!

PR #184, opened 2015-10-01
Consolidate member related admin endpoints

jwolski:
There's going to need to be a 'reuse' admin endpoint to undo member's that have been damped. At this point, it makes sense to start to combine all member-related admin endpoints. Thus, this PR.

@uber/ringpop


server/admin/member.js
```diff
@@ -58,4 +64,48 @@ module.exports = function createJoinHandler(ringpop) {
...
+        }
+
+        if (!ringpop.membership.localMember) {
+            process.nextTick(function() {
+                callback(errors.InvalidLocalMemberError());
+            });
+            return;
+        }
+
+        if (ringpop.membership.localMember.status === 'leave') {
```
benfleis: In other places we're codifying strings to objects, a la Status.Start. Should this also be codified?

jwolski: yes, it should. I'll fix.

PR #187, opened 2015-10-03
Remove listener after evaluating update to prevent leaking listeners

jwolski:
This bug'll cause leaky listeners and unnecessary CPU churn when redundant updates are processed.

@uber/ringpop


jwolski:
these are good suggestions.

PR #189, opened 2015-10-04
Add membership update types

jwolski:
This concretely defines Update types. Why this is need will make more sense in a PR still to come. In general, I think it's much better to explicitly define the shape of the updates that are transmitted as part of the gossip protocol.

@uber/ringpop


benfleis:
agreed, this definition is needed.

jwolski:
I killed BaseUpdate. My reason seemed better before you questioned its existence ;)

PR #194, opened 2015-10-04
WIP Membership-only leave w/ ttl

jwolski:
This PR supports a membership-only leave with a specifiable TTL on the leave operation. @Raynos asked about this on Friday. I've also taken a few liberties to clean up some piggyback code so that arbitrary options like `membershipOnly` can be disseminated. 

This code isn't ready for review. It's got no tests and there's still a fair bit of cleanup, but I've tested it locally and it does the right thing. I feel it's in a good enough place to let other people have at it in a test/stage environment.

Here's how you'd use it. Either invoke directly on the ringpop instance:

```
ringpop.leave({
  membershipOnly: true,
  ttl: 30000
});
```

or through its admin API which would look something like:

```
tcurl -p 192.168.1.10:3000 ringpop /admin/member/leave -3 '{"membershipOnly": false, "ttl": 30000}'
```

@uber/ringpop

### DIFF
--- a/index.js
+++ b/index.js
@@ -665,6 +665,13 @@ RingPop.prototype.handleOrProxyAll =
         }
     };
 
+// Valid opts for this function are those that are compatible with the
+// LeaveUpdate class defined in lib/membership/update.js.
+RingPop.prototype.leave = function leave(opts) {
+    this.membership.makeLeave(this.whoami(),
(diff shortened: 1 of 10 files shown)


Raynos:
Hmm; did this ever get implemented ?

jwolski:
It did not.

PR #195, opened 2015-10-04
Damping subprotocol

jwolski:
This PR introduces a Damper object which is responsible for:
* Tracking flappy members. A flappy member is one whose damp score exceeds the suppress limit.
* Periodically running a background damp timer that asks N-random other members for the damp scores of the same flappy members
* "Damping members" if the damp scores from the N-random members also exceed the suppress limit.
* Expiring damped members after members have been damped for greater than the suppression duration.

This PR does not actually 'treat' damped members other than logging an error when a Ringpop would have put a member in the damped status. The behavior to treat damped members will come in a later PR.

This PR is over 1200 lines. Take your time with it and of course, let me know if you have any high-level or fundamental questions.

Would like a lgtm from: @thanodnl @CorgiMan 
Would appreciate eyes and high-level feedback from: @benfleis @danielheller

@uber/ringpop

### DIFF
--- a/config.js
+++ b/config.js
@@ -68,10 +68,11 @@ Config.prototype._seed = function _seed(seed) {
 
     // Logger configs
     seedOrDefault('defaultLogLevel', LoggingLevels.info);
-    seedOrDefault('dampingLogLevel', LoggingLevels.debug);
+    seedOrDefault('dampingLogLevel', LoggingLevels.error);
     seedOrDefault('disseminationLogLevel', LoggingLevels.error);
-    seedOrDefault('gossipLogLevel', LoggingLevels.off);
+    seedOrDefault('gossipLogLevel', LoggingLevels.error);
     seedOrDefault('joinLogLevel', LoggingLevels.warn);
+    seedOrDefault('membershipLogLevel', LoggingLevels.error);
     seedOrDefault('suspicionLogLevel', LoggingLevels.error);
 
     // Gossip configs
@@ -83,6 +84,7 @@ Config.prototype._seed = function _seed(seed) {
     seedOrDefault('dampedMemberExpirationInterval', 60000);
     seedOrDefault('dampReqNVal', 6);
     seedOrDefault('dampReqRVal', 3);
+    seedOrDefault('dampReqTimeout', 1000);
     seedOrDefault('dampScoringEnabled', true);
     seedOrDefault('dampScoringDecayEnabled', true);
     seedOrDefault('dampScoringDecayInterval', 1000);
--- a/config.js
+++ b/config.js
@@ -88,7 +88,7 @@ Config.prototype._seed = function _seed(seed) {
     seedOrDefault('dampScoringEnabled', true);
     seedOrDefault('dampScoringDecayEnabled', true);
     seedOrDefault('dampScoringDecayInterval', 1000);
-    seedOrDefault('dampScoringHalfLife', 60);
+    seedOrDefault('dampScoringHalfLife', 60 * 5); // 5 minutes in seconds
     // TODO Initial should never be below min nor above max
     seedOrDefault('dampScoringInitial', 0);
     seedOrDefault('dampScoringMax', 10000);
--- a/config.js
+++ b/config.js
@@ -66,7 +66,7 @@ Config.prototype._seed = function _seed(seed) {
     // All config names should be camel-cased.
     seedOrDefault('TEST_KEY', 100, numValidator); // never remove, tests and lives depend on it
 
-    // Logger configs
+    // Logger configs; should be at least error by default.
     seedOrDefault('defaultLogLevel', LoggingLevels.info);
     seedOrDefault('dampingLogLevel', LoggingLevels.error);
     seedOrDefault('disseminationLogLevel', LoggingLevels.error);
--- a/config.js
+++ b/config.js
@@ -68,15 +68,21 @@ Config.prototype._seed = function _seed(seed) {
 
     // Logger configs
     seedOrDefault('defaultLogLevel', LoggingLevels.info);
-    seedOrDefault('dampingLogLevel', LoggingLevels.warn);
+    seedOrDefault('dampingLogLevel', LoggingLevels.debug);
+    seedOrDefault('disseminationLogLevel', LoggingLevels.error);
     seedOrDefault('gossipLogLevel', LoggingLevels.off);
     seedOrDefault('joinLogLevel', LoggingLevels.warn);
-    seedOrDefault('disseminationLogLevel', LoggingLevels.off);
+    seedOrDefault('suspicionLogLevel', LoggingLevels.error);
 
     // Gossip configs
     seedOrDefault('autoGossip', true);
 
     seedOrDefault('isCrossPlatform', false);
+    seedOrDefault('dampedErrorLoggingEnabled', false);
+    seedOrDefault('dampedMaxPercentage', 10);
+    seedOrDefault('dampedMemberExpirationInterval', 60000);
+    seedOrDefault('dampReqNVal', 6);
+    seedOrDefault('dampReqRVal', 3);
     seedOrDefault('dampScoringEnabled', true);
     seedOrDefault('dampScoringDecayEnabled', true);
     seedOrDefault('dampScoringDecayInterval', 1000);
@@ -89,6 +95,7 @@ Config.prototype._seed = function _seed(seed) {
     seedOrDefault('dampScoringReuseLimit', 2500);
     seedOrDefault('dampScoringSuppressDuration', 60 * 60 * 1000); // 1 hr in ms
     seedOrDefault('dampScoringSuppressLimit', 5000);
+    seedOrDefault('dampTimerInterval', 60 * 1000); // 1min in ms
 
     // Joiner config
     seedOrDefault('joinDelayMin', 100, numValidator); // ms
--- a/lib/gossip/damper.js
+++ b/lib/gossip/damper.js
@@ -210,7 +210,7 @@ Damper.prototype.expireDampedMembers = function expireDampedMembers() {
     var keys = Object.keys(this.dampedMembers);
     for (var i = 0; i < keys.length; i++) {
         var key = keys[i];
-        var memberDamping = this.dampedMembers[keys[i]];
+        var memberDamping = this.dampedMembers[key];
         var dampedDuration = this.Date.now() - memberDamping.timestamp;
         if (dampedDuration >= suppressDuration) {
             delete this.dampedMembers[key];
--- a/lib/gossip/damper.js
+++ b/lib/gossip/damper.js
@@ -155,6 +155,10 @@ Damper.prototype.addFlapper = function addFlapper(flapper) {
         Object.keys(this.flappers).length);
 
     if (this._getFlapperCount() === 1) {
+        this.logger.debug('ringpop damper starting damp timer; a flapper has been added', {
+            local: this.ringpop.whoami(),
+            flapper: flapper.address
+        });
         this._startDampTimer();
     }
 
@@ -188,6 +192,10 @@ Damper.prototype.dampMember = function dampMember(addr)  {
     this.addDampedMember(member.address);
 
     if (!this.isExpirationTimerEnabled) {
+        this.logger.debug('ringpop damper starting expiration timer; a member has been damped', {
+            local: this.ringpop.whoami(),
+            member: member.address
+        });
         this._startExpirationTimer();
     }
 
@@ -224,6 +232,9 @@ Damper.prototype.expireDampedMembers = function expireDampedMembers() {
 
     // If there are no more damped members, stop trying to expire them.
     if (Object.keys(this.dampedMembers).length === 0) {
+        this.logger.debug('ringpop damper stopping expiration timer; no damped members left', {
+            local: this.ringpop.whoami()
+        });
         this._stopExpirationTimer();
     }
 
@@ -349,9 +360,10 @@ Damper.prototype.initiateSubprotocol = function initiateSubprotocol(callback) {
 Damper.prototype.removeFlapper = function removeFlapper(flapper) {
     var address = flapper.address || flapper;
     if (!this.flappers[address]) {
-        this.logger.debug('ringpop flapper has not been added to the damper', {
+        this.logger.debug('ringpop damper did not remove flapper; it was not added. it may be damped.', {
             local: this.ringpop.whoami(),
-            flapper: address
+            flapper: address,
+            isDamped: !!this.dampedMembers[address]
         });
         return false;
     }
@@ -366,6 +378,9 @@ Damper.prototype.removeFlapper = function removeFlapper(flapper) {
         Object.keys(this.flappers).length);
 
     if (this._getFlapperCount() === 0) {
+        this.logger.debug('ringpop damper stopping damp timer; no flappers left', {
+            local: this.ringpop.whoami()
+        });
         this._stopDampTimer();
     }
 
@@ -380,7 +395,10 @@ Damper.prototype._fanoutDampReqs = function _fanoutDampReqs(flapperAddrs, dampRe
     for (var i = 0; i < dampReqMembers.length; i++) {
         var dampReqAddr = dampReqMembers[i].address;
         this.ringpop.stat('increment', 'damp-req.send');
-        this.ringpop.client.protocolDampReq(dampReqAddr, request,
+        this.ringpop.client.protocolDampReq({
+            host: dampReqAddr,
+            timeout: this.ringpop.config.get('dampReqTimeout')
+        }, request,
             createDampReqHandler(dampReqAddr));
     }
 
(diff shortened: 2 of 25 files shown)


config.js
```diff
@@ -68,14 +68,21 @@ Config.prototype._seed = function _seed(seed) {
 
     // Logger configs
     seedOrDefault('defaultLogLevel', LoggingLevels.info);
-    seedOrDefault('dampingLogLevel', LoggingLevels.warn);
+    seedOrDefault('dampingLogLevel', LoggingLevels.debug);
```
jwolski: TEMPORARYILY "debug".

lib/gossip/damper.js
```diff
@@ -0,0 +1,542 @@
...
+    // TODO Expire flappers?
+
+    this.flappers[flapper.address] = flapper;
+    this.logger.debug('ringpop damper added flapper', {
+        local: this.ringpop.whoami(),
+        flapper: flapper.address
+    });
+    this.ringpop.stat('increment', 'damper.flapper.added');
+    this.ringpop.stat('gauge', 'damper.flappers',
+        Object.keys(this.flappers.length));
```
thanodnl: This does not seem right, the keys of a number.

lib/gossip/damper.js
```diff
@@ -0,0 +1,542 @@
...
+    }
+};
+
+Damper.prototype.expireDampedMembers = function expireDampedMembers() {
+    var undampedMembers = [];
+    var suppressDuration = this.ringpop.config.get('dampScoringSuppressDuration');
+    var keys = Object.keys(this.dampedMembers);
+    for (var i = 0; i < keys.length; i++) {
+        var key = keys[i];
+        var memberDamping = this.dampedMembers[keys[i]];
```
thanodnl: use `key` instead of looking it up again.

lib/gossip/damper.js
```diff
@@ -0,0 +1,542 @@
...
+                timeDamped: dampedDuration
+            });
+        }
+    }
+
+    if (undampedMembers.length === 0) {
+        return undampedMembers;
+    }
+
+    this.emit('event', new events.DampedMemberExpirationEvent(undampedMembers));
```
thanodnl: `event` is not the best name here in my opinion. Can we be more specific what kind of event this is, eg `expired` ?

lib/gossip/damper.js
```diff
@@ -0,0 +1,542 @@
...
+        });
+    }
+
+    return true;
+
+    function scheduleExpiration() {
+        self.expirationTimer =
+            self.timers.setTimeout(function onTimeout() {
+                self.expireDampedMembers();
+                scheduleExpiration();
```
thanodnl: The call to `scheduleExpiration` regardless of the length of `dampedMembers`. This negates the call to `_stopExpirationTimer` on line 237.

lib/gossip/damper.js
```diff
@@ -0,0 +1,542 @@
...
+            });
+            self.emit('event', new events.DampReqFailedEvent(err));
+            callback();
+            return;
+        }
+
+        self.logger.debug('ringpop damper received all damp-req responses', {
+            local: self.ringpop.whoami(),
+            flappers: flapperAddrs,
+            responses: JSON.stringify(res)
```
thanodnl: Since this is in a debug log line I expect you don't want this logged in production, however you are still stringify-ing this in production.

lib/gossip/damper.js
```diff
@@ -0,0 +1,542 @@
...
+        });
+        var membersToDamp = Damper.getMembersToDamp(res, rVal, config);
+        if (membersToDamp.length === 0) {
+            self.ringpop.stat('increment', 'damper.damp-req.inconclusive');
+            self.logger.warn('ringpop damper inconclusive', {
```
thanodnl: Isn't this behavior the expected behavior in a completely health app?

lib/gossip/damper.js
```diff
@@ -0,0 +1,542 @@
...
+    }
+
+    delete this.flappers[address];
+    this.logger.debug('ringpop damper removed flapper', {
+        local: this.ringpop.whoami(),
+        flapper: address
+    });
+    this.ringpop.stat('increment', 'damper.flapper.removed');
+    this.ringpop.stat('gauge', 'damper.flappers',
+        Object.keys(this.flappers.length));
```
thanodnl: calling `keys` on a number/undefined. You probably want `Object.keys(this.flappers).length` here.

lib/gossip/damper.js
jwolski: Thanks yea. This stat was added recently. I caught it immediately and forgot to push up the fix.

lib/gossip/damper.js
jwolski: good catch.

lib/gossip/damper.js
jwolski: yes, I have been back and forth about how to issue these events. 

You're right to call this out. Lemme think about it.

lib/gossip/damper.js
jwolski: very nuanced! good thought though. I'll do something about it.

lib/gossip/damper.js
jwolski: Hm. Interesting question. Not necessarily, because if any node identifies a 'flapper' that means something "weird" is going on and the app isn't _completely_ healthy. But I understand what you mean about logging this at the warn level. `info` might be more appropriate.

lib/gossip/damper.js
jwolski: I think I understand what you're trying to get at, but since `expireDampedMembers()` is blocking/synchronous, the timer will either be cancelled/cleared by `_stopExpirationTimer`. Or am I misunderstanding?

thanodnl: As it looks to me the call on line 202 will invoke the function on 198, without checking to see if the expiration needs to be scheduled. Unless the `timers` module is doing some magic here. So even though the call above is synchronous it will immediately set the timer again after that function returns.

Further down there is a set of functions `_startDampTimer ` and `_stopDampTimer` which might be better to have for the expiration timer as well.

jwolski: Yes, this is true. But since we have single-threaded guarantees, we can ensure that if the timer were ever cancelled with `_stopExpirationTimer()`, then the interval would not tick and not subsequent `scheduleExpiration` would be called .

These same guarantees are not true for the dampTimer because the scheduling of its next interval is done after an asynchronous function completes, which in the time that it is running, could have been cancelled. That's why `dampTimerEnabled` exists for the dampTimer and not expirationtimer.

jwolski: UPDATE: I traced through the code and I see what you're pointing towards. `expireDampedMembers()` is calling `_stopExpirationTimer()` so at that point there would be no expiration timer to cancel.

Good catch. Fixing now.

config.js
```diff
@@ -68,14 +68,22 @@ Config.prototype._seed = function _seed(seed) {
 
     // Logger configs
     seedOrDefault('defaultLogLevel', LoggingLevels.info);
-    seedOrDefault('dampingLogLevel', LoggingLevels.warn);
+    seedOrDefault('dampingLogLevel', LoggingLevels.error);
     seedOrDefault('gossipLogLevel', LoggingLevels.off);
     seedOrDefault('joinLogLevel', LoggingLevels.warn);
+    seedOrDefault('suspicionLogLevel', LoggingLevels.off);
```
benfleis: Should all logging levels be at error, at a minimum? Even if they currently don't emit errors, this seems like the best default.

lib/gossip/damper.js
```diff
@@ -0,0 +1,580 @@
...
+    // since we're not actually damping them (marking the
+    // member's status to damped) at this point. Damped members
+    // are removed from this collection when the expiration timer
+    // removes them.
+    this.dampedMembers = {}; // indexed by address
+
+    // This damp timer runs only when there are flappy members
+    // add to the flappers collection above.
+    this.isDampTimerEnabled = false;
+    this.dampTimer = null;
```
benfleis: Would dampTimer ever be truthy, but Enabled be false? If not, should isDampTimerEnabled be a function?
Same question for expiration timer below.

lib/gossip/damper.js
```diff
@@ -0,0 +1,580 @@
...
+};
+
+Damper.prototype.expireDampedMembers = function expireDampedMembers() {
+    var undampedMembers = [];
+    var suppressDuration = this.ringpop.config.get('dampScoringSuppressDuration');
+    var keys = Object.keys(this.dampedMembers);
+    for (var i = 0; i < keys.length; i++) {
+        var key = keys[i];
+        var memberDamping = this.dampedMembers[key];
+        var dampedDuration = this.Date.now() - memberDamping.timestamp;
```
benfleis: this logic could be simplified by doing `var suppressTimestamp = this.Date.now() - suppressDuration;` outside the loop, and then a direct comparison `memberDamping.timestamp >= suppressTimestamp`.

lib/gossip/damper.js
benfleis: +1 on better naming -- agreed that if they're "strictly internal" that the naming ought to reflect that. I might suggest prepending w/ '_' a la "private" functions.

lib/gossip/damper.js
```diff
@@ -0,0 +1,580 @@
...
+    var self = this;
+    var config = this.ringpop.config;
+
+    // TODO Stop damp timer when cluster has reached damped limits
+    if (!this._validateDampedClusterLimits()) {
+        callback();
+        return;
+    }
+
+    // You may think that a check to verify that flappers actually
```
benfleis: This comment felt out of left field for me, and I realized afterward that it's because "initiateSubprotocol" doesn't tell me intuitively what this function does. If it was `initiateFanout(Subprotocol)`, maybe I would have had a clear notion of what's going on. Or a high level comment explaining it perhaps.

lib/gossip/damper.js
```diff
@@ -0,0 +1,580 @@
...
+            rVal: rVal,
+            nVal: nVal,
+            numDampReqMembers: dampReqMembers.length
+        });
+        this.emit('event', new events.DampReqUnsatisfiedEvent());
+        callback();
+        return;
+    }
+
+    this._fanoutDampReqs(flapperAddrs, dampReqMembers, rVal, onDampReqs);
```
benfleis: s/onDampReqs/onDampResps/ ?

lib/gossip/damper.js
```diff
@@ -0,0 +1,580 @@
...
+                errors: err
+            });
+            self.emit('event', new events.DampReqFailedEvent(err));
+            callback();
+            return;
+        }
+
+        // This log-level check is here to prevent stringification of responses
+        // in cases where level is set above debug.
+        if (self.logger.canLogAt(LoggingLevels.debug)) {
```
benfleis: netjes

lib/gossip/damper.js
```diff
@@ -0,0 +1,580 @@
...
+    var numPendingReqs = dampReqMembers.length;
+    var errors = [];
+    var results = [];
+
+    // Accumulate responses until rVal is satisfied or is impossible to satisfy because
+    // too many error responses.
+    function createDampReqHandler(addr) {
+        var start = Date.now();
+        return function onDampReq(err, res) {
+            var timing = Date.now() - start;
```
benfleis: `self.Date.now()` ?

lib/gossip/damper.js
```diff
@@ -0,0 +1,580 @@
...
+    }
+
+    var numPendingReqs = dampReqMembers.length;
+    var errors = [];
+    var results = [];
+
+    // Accumulate responses until rVal is satisfied or is impossible to satisfy because
+    // too many error responses.
+    function createDampReqHandler(addr) {
+        var start = Date.now();
```
benfleis: `self.Date.now()` ?

lib/gossip/damper.js
```diff
@@ -0,0 +1,580 @@
...
+        return function onDampReq(err, res) {
+            var timing = Date.now() - start;
+            self.ringpop.stat('timing', 'protocol.damp-req', timing);
+
+            // Prevents double-callback
+            if (typeof callback !== 'function') return;
```
benfleis: afaik, this violates or normal style guides. regardless of that, I would argue that return always deserves its own space, so that it's very clear that it's an exit from the current scope.

lib/logging/module_logger.js
```diff
@@ -47,6 +47,15 @@ ModuleLogger.prototype.trace = function trace(msg, meta) {
     this._log('trace', msg, meta);
 };
 
+// This function answers the question, "can I log at the desired log level
+// given the level configured for this logger?"
+ModuleLogger.prototype.canLogAt = function canLogAt(desiredLevel) {
```
benfleis: Love that this is added.

lib/membership/events.js
```diff
@@ -20,13 +20,24 @@
...
 
 function LocalMemberLeaveEvent(member, oldStatus) {
-    this.name = LocalMemberLeaveEvent.Name;
+    this.name = this.constructor.name;
     this.member = member;
     this.oldStatus = oldStatus;
 }
 
-LocalMemberLeaveEvent.Name = 'localMemberLeave';
+function ReusableEvent(member, oldDampScore) {
```
benfleis: What is reusable here? I don't understand the connection between reusable and an old damp score, and there's nothing to explain it.

Also, should all Damp events be prefixed with Damp[ing] or sth similar?

lib/membership/index.js
```diff
@@ -44,6 +46,17 @@ function Membership(opts) {
 
 util.inherits(Membership, EventEmitter);
 
+Membership.prototype.collectDampScores =
+        function collectDampScores(memberAddresses) {
+    var self = this;
+    return memberAddresses.reduce(function reduce(result, address) {
+        var member = self.findMemberByAddress(address);
+        if (!member) return result;
```
benfleis: could simplify and avoid short-circuit (with somewhat hidden return) by:
```
if (member) {
    result.push(...);
}
return result;
```

test/fixtures.js
```diff
@@ -0,0 +1,47 @@
...
+    opts = opts || {};
+    return new Member(ringpop, {
+        address: opts.address || '127.0.0.1:3001',
+        incarnationNumber: opts.incarnationNumber || Date.now(),
+        status: Member.Status.alive
+    });
+}
+
+function memberGenerator(ringpop) {
+    var basePort = 3001; // assums member with port 3000 is already in membership
```
benfleis: s/assums/assumes/

test/fixtures.js
```diff
@@ -0,0 +1,47 @@
...
+    var counter = 0;
+    return function genMember() {
+        return new Member(ringpop, {
+            address: '127.0.0.1:' + (basePort + counter++),
+            incarnationNumber: Date.now()
```
benfleis: this.Date.now()?  (i was assuming that the intention for that was testing)

benfleis:
Overall love the style and efforts for clarity. For me as a reader, there are a couple places that fall short of the high standard:
* rVal and nVal seem undernamed -- they tell me nothing by name, and I didn't see an introduction giving them context.
* IMHO some things should be prefixed with Damp/Damping, a la suppression events, reuse. If the intention is that suppress and reuse are "generic", I might also suggest more generic and clearly complementary (and idempotent) names, e.g., Ignore/Unignore, Exclude/Include, Embargo/Allow :), etc. (I didn't spend enough time to understand how generic they related funcs/code could be.)
* `if (foo) bar();` - this seems to violate what we've been doing thus far, as well as uber-standard.  (`grep -r 'if .*;' *.js lib test`)
* `request_response.js` doesn't feel 'top level', especially when accessing internals like ringpop.dissemination: -> lib/?

I am done.

config.js
jwolski: Good feedback. I think you're right.

lib/gossip/damper.js
jwolski: Good thought. Will fix.

lib/gossip/damper.js
jwolski: Ok. Will do.

lib/gossip/damper.js
jwolski: ack.

lib/gossip/damper.js
jwolski: ;)

lib/gossip/damper.js
jwolski: Yep, good catch. For completeness/correctness this should use `self.Date`.

lib/gossip/damper.js
jwolski: ack

lib/membership/events.js
jwolski: Reusable is not immediately obvious, particularly within the context of this PR because it's only ever fired, but not acted upon. I will add a comment. Good idea.

Re: the prefix. Maybe. I don't want to be overly verbose. Lemme chew.

lib/membership/index.js
jwolski: nice. will do.

test/fixtures.js
jwolski: thanks

test/fixtures.js
jwolski: It's not. It's just to give a real-looking incarnation number to test members.

lib/gossip/damper.js
jwolski: Ah, yea. I think this would be a good way to simplify.

lib/logging/module_logger.js
jwolski: wewt

PR #196, opened 2015-10-06
add tracing docs, and fix actual event struct

benfleis:
a bit of doc, and s/local/id/ on the struct. also added 'address' to the struct, future facing.

@wolski @CorgiMan

### DIFF
--- a/docs/architecture_design.md
+++ b/docs/architecture_design.md
@@ -84,6 +84,35 @@ Say the damp score for B exceeds the red line. A fans out a damp-req request to 
 ### Full Syncing
 Content coming soon...
 
+### Event Tracing
+
+Ringpop supports live event tracing, which can be sinked to either a TChannel endpoint, or a log-type entity. At present there is only one traceable event: `membership.checksum.update`.
+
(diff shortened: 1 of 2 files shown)


jwolski:
woot!

your tcat pull request has yet to be merged though. can you get that in so that the docs point to something available on mainline tcurl?

benfleis:
If that was possible, I would. I wrote tcat because I couldn't find anything already in existence. The docs refer to the branch, and I'll keep my fingers crossed. Otherwise I'll make my own tcat repo and put it there :/

jwolski:
:/ oh yes, you're right. I didnt originally see that you made mention of the `tcat` branch. That's fine until we have an alternate solution.

one last request, can you put the Event Tracing material under the How Ringpop Works section of "Architecture, Design, Implementation"? That's where I originally though that Feature descriptions would live. Eventually we might pull up a Features section to top-level, but right now, I don't think it makes sense to have Event Tracing be top-level by itself.

benfleis:
@jwolski wasn't sure if woot ~== lgtm, so I await a final :thumbsup:

jwolski:
LOL

PR #197, opened 2015-10-07
Use fingerprint32 over hash32.

thanodnl:
fingerprint is platform independent: https://github.com/lovell/farmhash#fingerprint.

Looks like TChannel ran into the same issue: https://github.com/uber/tchannel-node/commit/44a7edbb1facc9f53b29633820a4d7afbb03dc2d

/cc @uber/ringpop

### DIFF
--- a/config.js
+++ b/config.js
@@ -57,6 +57,7 @@ Config.prototype._seed = function _seed(seed) {
     // All config names should be camel-cased.
     seedOrDefault('TEST_KEY', 100); // never remove, tests and lives depend on it
     seedOrDefault('autoGossip', true);
+    seedOrDefault('isCrossPlatform', false);
     seedOrDefault('dampScoringEnabled', true);
     seedOrDefault('dampScoringDecayEnabled', true);
     seedOrDefault('dampScoringDecayInterval', 1000);
--- a/index.js
+++ b/index.js
@@ -35,6 +35,7 @@
 var _ = require('underscore');
 var EventEmitter = require('events').EventEmitter;
 var fs = require('fs');
+var farmhash = require('farmhash');
 var globalSetTimeout = require('timers').setTimeout;
 var hammock = require('uber-hammock');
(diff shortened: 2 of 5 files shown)


jwolski:
:+1:

thanodnl:
As discussed during the standup, this is a breaking change which we will wait to merge until there are other breaking changes to the protocol.

We have no reason to believe that the Hash32 is giving inconsistent results in different installations of node.

jwolski:
@thanodnl This'll be a breaking change, but I'd like to get out ahead of merging this in. How about doing something like the below?

Add in an `isCrossPlatform` option to the HashRing ctor...

```
new HashRing({
  isCrossPlatform: true
});
```

then...

```
function HashRing(opts) {
  this.isCrossPlatform = opts.isCrossPlatform;
  this.hashFunc = this.isCrossPlatform === true ? farmhash.fingerprint32 : farmhash.hash32;
```

jwolski:
oh, now I see it's not just the HashRing. We could do:

```
new Ringpop({
  isCrossPlatform: true
})
```

and in `config.js`:

```
seedOrDefault('isCrossPlatform', false);
```

and then introduce `lib/hash_func.js`:

```
module.exports = function createHashFunc(ringpop) {
  if (ringpop.config.get('isCrossPlatform')) {
    return farmhash.fingerprint32;
  } else {
    return farmhash.hash32;
  }
};
```

and then in HashRing and Membership:

```
var hashFunc = createHashFunc(ringpop);
hashFunc(...)
```

does that make sense?

thanodnl:
That makes total sense!

Then it will be merged as a non-breaking change, and when we have a breaking change coming up we can change the default behavior of ringpop so it uses

```
seedOrDefault('isCrossPlatform', true);
```

Will start integrating these changes in there.

jwolski:
looks awesome @thanodnl! Merging...

PR #198, opened 2015-10-09
Parse the key from the body of /admin/lookup.

thanodnl:
During the creation of the integration tests it came to my attention that the `/admin/lookup` endpoint was using the complete body as the key while the implementation in golang used the value provided in the `key` field of the body. This inconsistency made it impossible to test correctness of the lookup.

Looking in [`ringpop-admin`](https://github.com/uber/ringpop-admin/blob/master/lib/admin-client.js#L52-L54) (which I expect to be the only use of the lookup endpoint) revealed that sending a json object with a field `key` was the expected way to lookup a key.

This patch changes the `/admin/lookup/` endpoint in such a way that it will lookup the key for `your key` instead of `{key:"your key"}` when using `ringpop-admin lookup "host:port" --key "your key"`.

/cc @uber/ringpop

### DIFF
--- a/server/admin/lookup.js
+++ b/server/admin/lookup.js
@@ -19,11 +19,15 @@
 // THE SOFTWARE.
 'use strict';
 
+var safeParse = require('../../lib/util.js').safeParse;
+
 module.exports = function createLookupHandler(ringpop) {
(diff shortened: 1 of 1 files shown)


jwolski:
nice lgtm.

PR #199, opened 2015-10-09
Share arrays of added and removed addresses on ringChanged

kriskowal:
Over in Hyperbahn, I would like to incrementally update sorted arrays of affine relay addresses to reduce the cost of updating the service proxy in response to ring changes. To that end, I would like to thread the arrays of added and removed servers through the event.

I’ve added this feature and altered the existing test to verify the intended added and removed hosts, initializing the test to reflect reality since, but please check…it does not match my intuition.

r @uber/ringpop @jwolski

### DIFF
--- a/lib/on_membership_event.js
+++ b/lib/on_membership_event.js
@@ -21,7 +21,7 @@
 
 var Member = require('./membership/member.js');
 var MembershipEvents = require('./membership/events.js');
-var RingEvents = require('../ring/events.js');
+var RingEvents = require('./ring/events.js');
 
 function createChecksumComputedHandler(ringpop) {
     return function onMembershipChecksumComputed() {
--- a/lib/on_membership_event.js
+++ b/lib/on_membership_event.js
@@ -62,7 +62,8 @@ function createSetHandler(ringpop) {
         // efficiency gains when only having to compute the ring
         // checksum once.
         if (serversToAdd.length > 0) {
-            ringpop.ring.addRemoveServers(serversToAdd);
+            ringpop.ring.addRemoveServers(serversToAdd, []);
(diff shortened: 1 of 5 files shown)


lib/on_membership_event.js
```diff
@@ -62,7 +63,8 @@ function createSetHandler(ringpop) {
         // efficiency gains when only having to compute the ring
         // checksum once.
         if (serversToAdd.length > 0) {
-            ringpop.ring.addRemoveServers(serversToAdd);
+            ringpop.ring.addRemoveServers(serversToAdd, empty);
```
jcorbin: push down the var empty to here at least

Raynos:
This seems reasonable to me.

test/unit/index-test.js
```diff
@@ -394,18 +396,30 @@ test('emits ring changed event', function t(assert) {
...
+    }, {
+        adding: [],
+        removing: [node1Addr]
     });
 
     assertChanged(function assertIt() {
         ringpop.membership.makeAlive(node1Addr, magicIncNo);
+    }, {
+        adding: [],
+        removing: [node1Addr]
```
kriskowal: This formulation makes the tests pass, but doesn’t match my intuition of what should happen in this case.

kriskowal:
Fixed a bug and the test cases that pass make more sense now.

jwolski:
@kriskowal I pushed some commits on top of yours to keep it up-to-date with current approaches. I introduced `RingChangedEvent`. Make sure you take note of that. Otherwise, let's wait for a passing test and then I'll merge it in if it looks good to you.

kriskowal:
@jwolski This will get the job done. Thanks.

PR #200, opened 2015-10-09
Fix stuck requests

jwolski:
If we don't provide a Buffer as `res2` (as opposed to a String) to TChannel before sending the forwarded request back to the sender, the size of the buffer TChannel gives back (as `arg3`) is of an inconsistent size. This is particularly harmful when transferring Content-Length HTTP headers that use the size as a way of communicating how much data there is to consume. 

Before the fix was made, the test would hang waiting for data to arrive (as dictated by Content-Length). After the fix, the requestor no longer waits.

@uber/ringpop @jwyuan @charliezhang @Raynos @jcorbin


charliezhang:
lgtm

Raynos:
From reading this fix and test I suspect it's a tchannel core bug.

I suspect that fragmentation is broken when str.lenght !== buf.lenght

Cc @jcorbin

jwolski:
@Raynos I originally suspected that might be the case too. It'd be great to prove this is true. Let me know if/when you find something.

PR #201, opened 2015-10-13
Syncer

jwolski:
This feature takes full-syncing out of the ping and ping-req protocol and moves it to a periodic background timer run as often as `syncInterval`. Running the background syncer can also be disabled w/ config param `syncerEnabled`. 

Syncs requests will be made every 10s (by default). The response will contain the full state of the cluster if and only if the checksums of sender and receiver differ and there are no changes left to disseminate.

The reasoning: full syncs are a sort of anti-entropy mechanism and should be done under a more controlled manner than the gossip protocol. Subtle bugs in the gossip protocol can initiate more full syncs than desired and can cause more CPU and network burn.

Full syncing also supports gzip compression which is controlled by config `syncGzipEnabled`.

@uber/ringpop

### DIFF
--- a/client.js
+++ b/client.js
@@ -21,6 +21,7 @@
 
 var safeParse = require('./lib/util.js').safeParse;
 var TChannel = require('tchannel');
+var zlib = require('zlib');
 
 function RingpopClient(subChannel) {
     this.subChannel = subChannel;
@@ -35,27 +36,45 @@ function RingpopClient(subChannel) {
 }
 
 RingpopClient.prototype.adminConfigGet = function adminConfigGet(host, body, callback) {
-    this._request(host, '/admin/config/get', null, body, callback);
+    this._request(host, '/admin/config/get', null, body, null, callback);
 };
 
 RingpopClient.prototype.adminConfigSet = function adminConfigSet(host, body, callback) {
-    this._request(host, '/admin/config/set', null, body, callback);
+    this._request(host, '/admin/config/set', null, body, null, callback);
 };
 
 RingpopClient.prototype.adminGossipStart = function adminGossipStart(host, callback) {
-    this._request(host, '/admin/gossip/start', null, null, callback);
+    this._request(host, '/admin/gossip/start', null, null, null, callback);
 };
 
 RingpopClient.prototype.adminGossipStop = function adminGossipStop(host, callback) {
-    this._request(host, '/admin/gossip/stop', null, null, callback);
+    this._request(host, '/admin/gossip/stop', null, null, null, callback);
 };
 
 RingpopClient.prototype.adminGossipTick = function adminGossipTick(host, callback) {
-    this._request(host, '/admin/gossip/tick', null, null, callback);
+    this._request(host, '/admin/gossip/tick', null, null, null, callback);
 };
 
-RingpopClient.prototype.protocolSync = function protocolSync(host, body, callback) {
-    this._request(host, '/protocol/sync', null, body, callback);
+RingpopClient.prototype.protocolSync = function protocolSync(opts, callback) {
+    opts = opts || {};
+    var host = opts.host;
+    var body = opts.body;
+
+    if (!host) {
+        process.nextTick(function onTick() {
+            callback(new Error('Bad request: host is required'));
+        });
+        return;
+    }
+
+    if (!body) {
+        process.nextTick(function onTick() {
+            callback(new Error('Bad request: body is required'));
+        });
+        return;
+    }
+
+    this._request(host, '/protocol/sync', null, body, opts, callback);
 };
 
 RingpopClient.prototype.destroy = function destroy(callback) {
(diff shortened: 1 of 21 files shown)


client.js
```diff
@@ -90,6 +116,18 @@ RingpopClient.prototype._request = function _request(host, endpoint, head, body,
             return;
         }
 
+        if (opts.gzip === true) {
```
thanodnl: It would make more sense to me if the headers of the response would indicate that the body is gzipped.

With the current PR both nodes involved in a sync would need to have the same configuration which could cause strange issues in production one day. By indicating gzippedness in the header of the response the client can deduct on its own that it needs to unzip the contents before parsing it.

jwolski: u r completely right. great idea!

lib/gossip/dissemination.js
```diff
@@ -22,13 +22,14 @@
...
 
+var EMPTY = [];
 var LOG_10 = Math.log(10);
 
 function Dissemination(ringpop) {
     this.ringpop = ringpop;
     this.ringpop.on('ringChanged', this.onRingChanged.bind(this));
 
-    this.changes = {};
+    this.membershipChanges = {};
```
jwolski: NOTES FOR REVIEWER: Renamed var to be more explicit about what type of changes are being disseminated.

lib/gossip/index.js
```diff
@@ -94,7 +94,6 @@ Gossip.prototype.start = function start() {
         return;
     }
 
-    this.ringpop.membership.shuffle();
```
jwolski: NOTES FOR REVIEWER: Shuffling is done lazily by the MembershipIterator now.

client.js
```diff
@@ -90,6 +95,18 @@ RingpopClient.prototype._request = function _request(host, endpoint, head, body,
             return;
         }
 
+        if (head && head.gzip === true) {
```
thanodnl: `head` in this context is the header of the request, not of the response. In the case you are making a gzipped request to an older version (without gzip support) it will send you the plain information back.

I do think it is also good to signal in the request header that you want the response to be gzipped, but lets confirm in the response header that it is actually gzipped before trying to unzip it. Which involves adding that header I believe: https://github.com/uber/ringpop-node/blob/syncer/server/protocol/sync.js#L63

lib/gossip/index.js
thanodnl: Now that the membership iterator is doing this lazy shuffle one of the characteristics of SWIM as described in `4.3. Round-Robin Probe Target Selection: Providing
Time-Bounded Strong Completeness` is not applicable any more.

In the paper they say "a newly joining member is inserted
in the membership list at a position that is chosen uniformly
at random". Now a newly joined member will only have a chance of being pinged when 1 node reaches the end of its pingable members and reshuffles the membership list into the members array of the MembershipIterator.

client.js
jwolski: Good catch. Will fix.

lib/gossip/index.js
jwolski: Syncer makes a new iterator every interval: https://github.com/uber/ringpop-node/pull/201/files#diff-79519da59c526a4c3df3c9edfd70adc7R105

thanodnl: It is good that the ping mechanism and the syncer use different iterators, but as discussed the iterator should slightly change so that it inserts newly added nodes at a random position in the  MembershipIterator.members to conform to chapter 4.3 of the swim paper again.

Before the changes to the lazy shuffle mechanism the line at https://github.com/uber/ringpop-node/blob/syncer/lib/membership/index.js#L285 made sure that newly added nodes have a fair chance of being pinged by already existing nodes, now that the list only gets shuffled when exhausted this fairness went away.

I think it can be solved by making the MembershipIterator listen `updated` events on the membership. But maybe it is easier to add a new member event in this block: https://github.com/uber/ringpop-node/blob/syncer/lib/membership/index.js#L277-L294 since you specifically want to add only new members to the list of the iterator.

lib/gossip/syncer.js
```diff
@@ -0,0 +1,192 @@
...
+                local: self.ringpop.whoami(),
+                target: pingableMember.address,
+                currentChecksum: self.ringpop.membership.checksum,
+                targetChecksum: response.membershipChecksum
+            });
+            cleanup(new events.SyncEmptyEvent());
+            return;
+        }
+
+        var updates = self.ringpop.membership.update(membershipChanges);
```
thanodnl: It would be good to check the checksum after all received changes are applied, if they are still a mismatch it might be worth to send this nodes membership list to the other party to get them applied.

We have seen situations where there was one (or more) nodes that thought they were the complete ring. They will never initiate a sync to any other node of the actual ring, and thus never get their membership list.

If the ring would do a sync to such a node it will find that after applying that nodes changes they are still in an inconsistent state. Sending this nodes members to the single-ring node would fix such a situation.

jwolski: Agreed, but I'm going to punt on this for this PR. It's an optimization that we can add later.

PR #204, opened 2015-10-19
Join duration w/ backoff

jwolski:
This is a quick hotfix to apply a backoff delay for the join process.

It'll backoff exponentially up to the max join delay (2 mins) in between each retry and try to join up to the max join duration (20 mins). Ringpop will flag the developer with an error when the join delay reaches its max. If join duration is exceeded, Ringpop'll still do nothing and just idle naively waiting for something to be done. A continuous joiner and refusing gossip communication are left for future PRs.

@uber/ringpop

### DIFF
--- a/config.js
+++ b/config.js
@@ -82,7 +82,12 @@ Config.prototype._seed = function _seed(seed) {
     seedOrDefault('dampScoringReuseLimit', 2500);
     seedOrDefault('dampScoringSuppressDuration', 60 * 60 * 1000); // 1 hr in ms
     seedOrDefault('dampScoringSuppressLimit', 5000);
-    seedOrDefault('maxJoinAttempts', 50, numValidator);
+
+    // Joiner config
+    seedOrDefault('joinDelayMin', 100, numValidator); // ms
+    seedOrDefault('joinDelayMax', 2 * 60 * 1000, numValidator); // 2 min in ms
+    seedOrDefault('maxJoinDuration', 20 * 60 * 1000, numValidator); // 20 mins in ms
+
     seedOrDefault('memberBlacklist', [], function validator(vals) {
         return _.all(vals, function all(val) {
             return val instanceof RegExp;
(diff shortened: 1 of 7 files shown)


config.js
```diff
@@ -70,7 +70,13 @@ Config.prototype._seed = function _seed(seed) {
     seedOrDefault('dampScoringReuseLimit', 2500);
     seedOrDefault('dampScoringSuppressDuration', 60 * 60 * 1000); // 1 hr in ms
     seedOrDefault('dampScoringSuppressLimit', 5000);
-    seedOrDefault('maxJoinAttempts', 50, numValidator);
+
+    // Joiner config
+    seedOrDefault('joinDelayMin', 100, numValidator); // ms
+    seedOrDefault('joinDelayMax', 2 * 60 * 1000, numValidator); // 2 min in ms
+    seedOrDefault('joinTroubleErrorEnabled', true);
+    seedOrDefault('maxJoinDuration', 20 * 60 * 1000, numValidator); // 20 mins in ms
```
benfleis: suggestion: joinDurationMax -- thus join* + same as above pattern w/ joinDelayMax

lib/gossip/joiner.js
```diff
@@ -302,31 +258,67 @@ JoinCluster.prototype.join = function join(callback) {
...
+            var oldJoinDelay = self.joinDelay;
+            var delayMax = self.ringpop.config.get('joinDelayMax');
+            var delayMin = self.ringpop.config.get('joinDelayMin');
+            var newDelay = delayMin * Math.pow(2, self.joinRetries);
+            var withFuzz = Math.floor(Math.random() *
+                ((newDelay * 1.5) - newDelay)) + newDelay;
+            self.joinDelay = Math.min(delayMax, withFuzz);
+
+            // Determine if the join delay has exceeded the maximum
+            // delay and send out a warning letting developer's know
```
benfleis: s/er's/ers/

benfleis:
nits aside, lgtm

lib/gossip/joiner.js
jwolski: thx

config.js
jwolski: agreed, but I did this to preserve backwards compatibility as there is an option that could have been provided to Ringpop in previous releases that would have set the maxJoinDuration. Keeping it named as such will retain that value and not break existing apps.

thanodnl:
:+1: lgtm

PR #205, opened 2015-10-20
No debug flags

jwolski:
I was working on a PR that refused gossip protocol communication until Ringpop is ready and wound up getting mad that the debug clear/set code is still around. I wanted to clean up the log sites that depended on it and continue to reduce the Ringpop prototype interface so I did this work here instead of trying to do two different things in that PR.

Debug flags have historically been used to toggle gossip (or protocol) logging from tick-cluster. I'm now trying to generalize 'debug flags' or more accurately, toggling gossip protocol logging to be supported by the in-memory config object. With this change we'll be able to turn on gossip logging over the /admin/config/set and /admin/config/get endpoints.

@uber/ringpop


benfleis:
lgtm

PR #206, opened 2015-10-21
Module loggers w/ individually configurable levels

jwolski:
Some apps like to log. Some apps don't. Some logs are nice to have for debugging. Some logs aren't. Module loggers allow different Ringpop components/modules to have individually configurable logging levels which are settable in the Ringpop constructor or over the /admin/config/set endpoints.

reviewers: @thanodnl @benfleis 
cc: @uber/ringpop

### DIFF
--- a/config.js
+++ b/config.js
@@ -21,6 +21,7 @@
 
 var _ = require('underscore');
 var EventEmitter = require('events').EventEmitter;
+var LoggingLevels = require('./lib/logging/levels.js');
 var util = require('util');
 
 // This Config class is meant to be a central store
@@ -65,9 +66,13 @@ Config.prototype._seed = function _seed(seed) {
     // All config names should be camel-cased.
     seedOrDefault('TEST_KEY', 100, numValidator); // never remove, tests and lives depend on it
 
+    // Logger configs
+    seedOrDefault('defaultLogLevel', LoggingLevels.info);
+    seedOrDefault('gossipLogLevel', LoggingLevels.off);
+    seedOrDefault('joinerLogLevel', LoggingLevels.warn);
+
     // Gossip configs
     seedOrDefault('autoGossip', true);
-    seedOrDefault('gossipLogLevel', Config.Defaults.gossipLogLevel);
 
     seedOrDefault('isCrossPlatform', false);
     seedOrDefault('dampScoringEnabled', true);
--- a/config.js
+++ b/config.js
@@ -69,7 +69,7 @@ Config.prototype._seed = function _seed(seed) {
     // Logger configs
     seedOrDefault('defaultLogLevel', LoggingLevels.info);
     seedOrDefault('gossipLogLevel', LoggingLevels.off);
-    seedOrDefault('joinerLogLevel', LoggingLevels.warn);
+    seedOrDefault('joinLogLevel', LoggingLevels.warn);
 
     // Gossip configs
     seedOrDefault('autoGossip', true);
--- a/lib/gossip/ping-req-sender.js
+++ b/lib/gossip/ping-req-sender.js
@@ -19,7 +19,6 @@
 // THE SOFTWARE.
 'use strict';
 
-var gossipLogger = require('../loggers.js').gossipLogger;
 var safeParse = require('../util').safeParse;
 var TypedError = require('error/typed');
 
@@ -52,6 +51,7 @@ var PingReqPingError = TypedError({
 
 function PingReqSender(ring, member, target, callback) {
     this.ring = ring;
+    this.gossipLogger = this.ring.loggerFactory.getLogger('gossip');
     this.member = member;
     this.target = target;
     this.callback = callback;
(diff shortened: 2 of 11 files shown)


lib/logging/module_logger.js
```diff
@@ -0,0 +1,60 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+'use strict';
+
+var Levels = require('./level.js');
```
thanodnl: levels.js

jwolski: thx

lib/gossip/ping-req-sender.js
```diff
@@ -184,7 +184,7 @@ module.exports = function sendPingReq(opts, callback) {
     for (var i = 0; i < pingReqMembers.length; i++) {
         var pingReqMember = pingReqMembers[i];
 
-        gossipLogger(ringpop, 'ringpop ping-req send', {
+        this.gossipLogger.info('ringpop ping-req send', {
```
thanodnl: This is outside of the scope of the `PingReqSender`

jwolski: thx. u r correct. i will fix.

config.js
```diff
@@ -65,9 +66,13 @@ Config.prototype._seed = function _seed(seed) {
     // All config names should be camel-cased.
     seedOrDefault('TEST_KEY', 100, numValidator); // never remove, tests and lives depend on it
 
+    // Logger configs
+    seedOrDefault('defaultLogLevel', LoggingLevels.info);
+    seedOrDefault('gossipLogLevel', LoggingLevels.off);
+    seedOrDefault('joinerLogLevel', LoggingLevels.warn);
```
benfleis: reading 'joiner' makes me wonder -- is level setting intended to apply to objects (Joiner) or concepts (join)?

if the latter, then perhaps it should be s/joiner/join/ to match other join knobs?

lib/logging/levels.js
```diff
@@ -0,0 +1,56 @@
...
+var levels = ['trace', 'debug', 'info', 'warn', 'error', 'off'];
+
+// e.g. trace => 'trace', debug => 'debug', etc.
+var namedLevels = levels.reduce(function each(res, level) {
+    res[level] = level;
+    return res;
+}, {});
+
+// e.g. trace => 0, debug => 1, etc.
+var levelToInt = {};
```
benfleis: Nit: IMHO there is value in corresponding to [syslog standards](https://en.wikipedia.org/wiki/Syslog#Severity_level) for numbering. Maybe it never matters due to export, but my €0,02.

lib/logging/logger_factory.js
```diff
@@ -19,15 +19,33 @@
...
+        case 'gossip':
+            this.loggers[name] = new ModuleLogger(this.ringpop,
+                'gossipLogLevel');
+            break;
+        case 'joiner':
+            this.loggers[name] = new ModuleLogger(this.ringpop,
+                'joinerLogLevel');
+            break;
+        default:
+            this.loggers[name] = new ModuleLogger(this.ringpop,
```
benfleis: this is boss for adding temporary/local logging, and still being able to control it :)

benfleis: on second read (and reading further down), could this be better handled as a fallthrough, so that I can set foobarLogLevel in my config, and have it still work?

benfleis:
added a couple thoughts, trlgtm.

config.js
jwolski: fair. 'packages' or 'modules' is a good delineation. will fix.

lib/logging/levels.js
jwolski: interesting. i approve.

lib/logging/logger_factory.js
jwolski: lemme think about this.

lib/logging/levels.js
jwolski: on second thought, while it's nice to reflect some established standards, these int values are completely hidden by Ringpop. There are log levels that dont really apply to the logger Ringpop depends on e.g. critical, notice, emergency. I'm going to leave as is.

lib/logging/logger_factory.js
jwolski: i thought about it and will keep as is until we require such flexibility.

PR #207, opened 2015-10-22
Track number of inflight forwarded requests

jwolski:
Add stats to track the number of inflight forwarded requests.

@uber/ringpop


benfleis:
lgtm

PR #208, opened 2015-10-22
Backpressure at routing egress/ingress

jwolski:
This pull request applies backpressure at Ringpop's forwarding egress and ingress by using the number of inflight requests and the current event loop lag as the metric to use when evaluating how "overwhelmed" an application is.

reviewers: @thanodnl @danielheller
cc: @uber/ringpop @weikai77 @markyen @fuj

### DIFF
--- a/config.js
+++ b/config.js
@@ -94,6 +94,12 @@ Config.prototype._seed = function _seed(seed) {
     seedOrDefault('joinTroubleErrorEnabled', true);
     seedOrDefault('maxJoinDuration', 20 * 60 * 1000, numValidator); // 20 mins in ms
 
+    // Forwarding config
+    seedOrDefault('backpressureEnabled', false);
+    seedOrDefault('lagSamplerInterval', 1000, numValidator);
+    seedOrDefault('maxEventLoopLag', 1500, numValidator);
+    seedOrDefault('maxInflightRequests', 5000, numValidator);
+
     seedOrDefault('memberBlacklist', [], function validator(vals) {
         return _.all(vals, function all(val) {
             return val instanceof RegExp;
--- a/lib/request-proxy/index.js
+++ b/lib/request-proxy/index.js
@@ -19,17 +19,21 @@
 // THE SOFTWARE.
 'use strict';
 
-// 3rd party dependencies
 var body = require('body');
 var hammock = require('uber-hammock');
-var PassThrough = require('stream').PassThrough;
-var TypedError = require('error/typed');
-
-// 1st party dependencies
 var numOrDefault = require('../util.js').numOrDefault;
+var PassThrough = require('stream').PassThrough;
 var safeParse = require('../util.js').safeParse;
 var sendRequest = require('./send.js');
+var TypedError = require('error/typed');
 
+var HTTP_TOO_MANY_REQUESTS = 429;
+// Very very slight optimization, but there's no reason
+// to stringify the headers for every ingress request when
+// backpressure is applied. Once is enough.
+var INGRESS_BACKPRESSURE_HEAD = JSON.stringify({
+    statusCode: HTTP_TOO_MANY_REQUESTS
+});
 var InvalidCheckSumError = TypedError({
     type: 'ringpop.request-proxy.invalid-checksum',
     message: 'Expected the remote checksum to match local ' +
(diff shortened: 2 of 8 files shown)


config.js
```diff
@@ -89,6 +89,9 @@ Config.prototype._seed = function _seed(seed) {
     seedOrDefault('joinTroubleErrorEnabled', true);
     seedOrDefault('maxJoinDuration', 20 * 60 * 1000, numValidator); // 20 mins in ms
 
+    // Forwarding config
+    seedOrDefault('maxInflightRequests', 1000, numValidator);
```
thanodnl: Is there a way to turn this off? To switch back to how ringpop is currently behaving?

jwolski: No, I thought to add a on/off switch.

lib/request-proxy/index.js
```diff
@@ -230,6 +256,75 @@ proto.handleRequest = function handleRequest(head, body, cb) {
...
+    var backpressureEnabled = config.get('backpressureEnabled');
+    var max = config.get('maxEventLoopLag');
+    return backpressureEnabled && this.ringpop.lagSampler.currentLag > max;
+};
+
+RequestProxy.prototype._shouldApplyBackpressure =
+    function _shouldApplyBackpressure(egressOrIngress, callback) {
+    if (this._hasExceededMaxInflightRequests()) {
+        this.ringpop.stat('increment', 'requestProxy.refused.inflight');
+        callback(new TooManyRequestsError({
```
danielheller: nextTick() or setImmediate() seems appropriate here, since other invokers of the callback are likely to do so async.

package.json
```diff
@@ -34,6 +34,7 @@
     "farmhash": "^0.2.0",
     "metrics": "^0.1.8",
     "node-uuid": "^1.4.3",
+    "toobusy": "^0.2.4",
```
danielheller: Good stuff, thumbs up to @benfleis.

lib/request-proxy/index.js
```diff
@@ -197,18 +219,22 @@ proto.handleRequest = function handleRequest(head, body, cb) {
...
+    if (this._shouldApplyBackpressure('ingress', function onApply(err) {
+        cb(err);
+    })) {
+        return;
+    }
 
-    ringpop.logger.trace('handleRequest emitting request', {
-        url: url
-    });
+    this._incrementInflightRequests();
```
danielheller: Am I mistaken or does doing this accounting here mean that we count requests both coming and going?  I'm not saying that's wrong, but wouldn't it have the interesting effect that if requests are evenly distributed across the cluster, our in-flight limit in practice is 1/2 of the nominal value?  Is this intentional?

danielheller: I guess this makes sense, since requests do cost us both coming and going.

config.js
```diff
@@ -83,12 +83,20 @@ Config.prototype._seed = function _seed(seed) {
...
     // Joiner config
     seedOrDefault('joinDelayMin', 100, numValidator); // ms
     seedOrDefault('joinDelayMax', 2 * 60 * 1000, numValidator); // 2 min in ms
     seedOrDefault('joinTroubleErrorEnabled', true);
     seedOrDefault('maxJoinDuration', 20 * 60 * 1000, numValidator); // 20 mins in ms
 
+    // Forwarding config
+    seedOrDefault('backpressureEnabled', true);
+    seedOrDefault('maxEventLoopLag', 1500, numValidator);
+    seedOrDefault('maxInflightRequests', 1000, numValidator);
```
danielheller: Does the back of an envelope support these defaults?  The lag makes intuitive sense to me.  I don't have a strong intuition about max inflight requests and haven't seen stats myself, though 1000 is certainly plausible.

lib/request-proxy/index.js
```diff
@@ -49,22 +56,27 @@ function RequestProxy(opts) {
...
+    // have not yet been added to the sends collection. We maintain
+    // a separate counter in numInflightRequests to make sure that
+    // even requests-to-be-sent are accounted for.
+    this.numInflightRequests = 0;
 }
 
-var proto = RequestProxy.prototype;
-
-proto.removeSend = function removeSend(send) {
+RequestProxy.prototype.removeSend = function removeSend(send) {
```
danielheller: Like the cleanup.

danielheller:
lgtm modulo a couple of minor questions!

config.js
jwolski: It's a fair point. I had the intention to be fairly conservative with these numbers so as not to disrupt current expectations by default. But, even 1k might be too low. Let me revisit.

lib/request-proxy/index.js
jwolski: werd

lib/request-proxy/index.js
jwolski: you got it.

lib/request-proxy/index.js
jwolski: you're probably right. let me revisit the flow and make sure nextTick is/isn't appropriate here.

package.json
jwolski: :+1:

PR #209, opened 2015-10-23
Stat number of retries performed during join process

jwolski:
This stat will give us a better idea about how successful the join fan-out is for all Ringpop-based applications.

@uber/ringpop


benfleis:
ship it

jwolski:
wewt

PR #211, opened 2015-10-28
Ingress/egress counters

jwolski:
Stats for ingress and egress requests so we can get individual per sec rates.

@uber/ringpop


danielheller:
lgtm

PR #212, opened 2015-10-28
Use the correct isPinging flags

jwolski:
`this.ringpop.isPinging` was never set to true. 

`this.ringpop.gossip.isPinging` was never set to false.

At some point when I rearranged code, I'm sure I unintentionally broke this. It likely isn't showing up as any sort of production bug because it was originally put in place to make sure that tick-cluster's manual tick never ticked while a gossip tick was outstanding.

@uber/ringpop

### DIFF
--- a/lib/gossip/index.js
+++ b/lib/gossip/index.js
@@ -28,6 +28,7 @@ function Gossip(options) {
     this.minProtocolPeriod = options.minProtocolPeriod ||
         Gossip.Defaults.minProtocolPeriod;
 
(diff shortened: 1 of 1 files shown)


thanodnl:
I would suggest setting `isPinging` explicitly to `false` in the constructor of `Gossip`. Although the magic of javascript and falsy value's make this currently work.

jwolski:
yes. good idea @thanodnl

thanodnl:
lgmt

PR #213, opened 2015-11-01
Add a damping logger for flap damping logging

jwolski:
Easy one. Change damp score logging so that its controlled by module loggers.

@uber/ringpop


benfleis:
lgtm

PR #218, opened 2015-11-12
Don't disseminate join list

CorgiMan:
@thanodnl @jwolski

### DIFF
--- a/lib/gossip/dissemination.js
+++ b/lib/gossip/dissemination.js
@@ -132,6 +132,14 @@ Dissemination.prototype.resetMaxPiggybackCount = function resetMaxPiggybackCount
     this.maxPiggybackCount = Dissemination.Defaults.maxPiggybackCount;
 };
 
+Dissemination.prototype.getChangeByAddress = function getChangeByAddress(address) {
+    return this.changes[address];
+};
+
+Dissemination.prototype.getChangesCount = function getChangesCount() {
+    return Object.keys(this.changes).length;
+};
+
 Dissemination.prototype._issueAs = function _issueAs(filterChange, mapChanges) {
     var changesToDisseminate = [];
 
--- a/lib/on_membership_event.js
+++ b/lib/on_membership_event.js
@@ -54,8 +54,6 @@ function createSetHandler(ringpop) {
                 serversToAdd.push(update.address);
                 ringpop.suspicion.start(update);
             }
-
-            ringpop.dissemination.recordChange(update);
         }
 
         // Must add/remove servers from ring in batch. There are
--- a/test/integration/join-test.js
+++ b/test/integration/join-test.js
@@ -27,7 +27,7 @@ var testRingpopCluster = require('../lib/test-ringpop-cluster.js');
 test('bootstrap with self is ok', function t(assert) {
     var ringpop = allocRingpop();
 
-    ringpop.bootstrap([ringpop.hostPort], onBootstrap);
+    ringpop.bootstrap([ringpop.whoami()], onBootstrap);
 
     function onBootstrap(err) {
         assert.ifError(err);
@@ -40,7 +40,7 @@ testRingpopCluster({
     size: 1,
     waitForConvergence: false
 }, 'one node can join', function t(bootRes, cluster, assert) {
-    assert.ifErr(bootRes[cluster[0].hostPort].err, 'no error occurred');
+    assert.ifErr(bootRes[cluster[0].whoami()].err, 'no error occurred');
     assert.end();
 });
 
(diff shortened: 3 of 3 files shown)


jwolski:
please write some unit tests for this change

lib/on_membership_event.js
```diff
@@ -55,7 +55,9 @@ function createSetHandler(ringpop) {
                 ringpop.suspicion.start(update);
             }
 
-            ringpop.dissemination.recordChange(update);
+            if(update.address === ringpop.whoami()) {
```
jwolski: space after if: `if (`

jwolski: Also.. I actually don't think this is necessary. I believe that the 'local' member is added to the dissemination component before the membership set event is fired. Therefore, we won't need to `recordChange()` at all in the event handler. We might be able to kill this line altogether and achieve the same.

lib/on_membership_event.js
```diff
@@ -54,8 +54,6 @@ function createSetHandler(ringpop) {
                 serversToAdd.push(update.address);
                 ringpop.suspicion.start(update);
             }
```
CorgiMan: You were right @jwolski we can just remove this line

CorgiMan:
Will add unit test. Have added integration test here:
https://github.com/uber/ringpop-common/pull/18

thanodnl:
<3 1 line PR's that save tons of bandwidth :).

Did you look at how `ringpop-go` behaves? And is that one still passing the integration tests after this change?

test/integration/join-test.js
```diff
@@ -108,6 +108,22 @@ testRingpopCluster({
 });
 
 testRingpopCluster({
+    tap: function tap(cluster) {
+        cluster.forEach(function eachNode(node) {
```
jwolski: Put a comment here about why disabling gossip is necessary.

test/integration/join-test.js
```diff
@@ -108,6 +108,22 @@ testRingpopCluster({
...
+            node.gossip.start = function noop() {};
+        });
+    },
+    waitForConvergence: false
+}, 'do not disseminate join list', function t(bootRes, cluster, assert) {
+    cluster.forEach(function eachNode(node) {
+        var changesKeys = Object.keys(node.dissemination.changes);
+        assert.equal(changesKeys.length, 1, 'only one change to be disseminated');
+        assert.equal(changesKeys[0], node.hostPort, 'disseminate only yourself');
```
jwolski: instead of `node.hostPort` use `node.whoami()`;

test/integration/join-test.js
```diff
@@ -108,6 +108,22 @@ testRingpopCluster({
...
 
 testRingpopCluster({
+    tap: function tap(cluster) {
+        cluster.forEach(function eachNode(node) {
+            node.gossip.start = function noop() {};
+        });
+    },
+    waitForConvergence: false
+}, 'do not disseminate join list', function t(bootRes, cluster, assert) {
+    cluster.forEach(function eachNode(node) {
```
jwolski: call `assert.plan(X)` to make sure that the "right" number of assertions are happening. I suggest this because for some reason if cluster were empty, forEach wouldnt iterate over anything and the test would still pass.

test/integration/join-test.js
```diff
@@ -108,6 +108,22 @@ testRingpopCluster({
...
+        var changesKeys = Object.keys(node.dissemination.changes);
```
jwolski: Instead of reaching into the internal state of the dissemination object and pulling out the changes, expose an API so you can call:

```
dissemination.getChangesCount();
dissemination.getChangeAt(0);
```

This helps protect having to modify tests if we were rename the member variable of the dissemination object or refactor in some other way.

lib/gossip/dissemination.js
```diff
@@ -131,6 +131,14 @@ Dissemination.prototype.resetMaxPiggybackCount = function resetMaxPiggybackCount
     this.maxPiggybackCount = Dissemination.Defaults.maxPiggybackCount;
 };
 
+Dissemination.prototype.getChangeByAddress = function getChangeByAddress(address) {
+    return this.changes[address]
```
jwolski: missing semicolon

test/integration/join-test.js
```diff
@@ -108,6 +108,21 @@ testRingpopCluster({
 });
 
 testRingpopCluster({
+    // We dissable gossip because we don't want to exhaust the changes through pings
```
jwolski: typo -> disable

jwolski:
lgtm. let's test w/ ringpopd before merging.

CorgiMan:
Nice, good idea, thnx for review, I'll keep you updated

jwolski:
@CorgiMan can you rebase onto latest master and merge conflicts?

CorgiMan:
@jwolski Rebased. I had to force it for a reason unknown to me

thanodnl:
Rebasing always needs a force push if the branch was already pushed to the
remote since it can't forward the ref to the new location on the tree. It
cannot forward since the ancestors of the commit are different.

-- Nils

CorgiMan:
@dansimau ^

thanodnl:
All but one commit from this PR are now in master as it seems, don't we need: https://github.com/uber/ringpop-node/commit/286e5d9d36dcf7217560392644563c39fc6f5b1f in master as well?

PR #223, opened 2015-11-26
Log membership and ring checksum computed events

jwolski:
Helps us to measure convergence

@uber/ringpop

### DIFF
--- a/config.js
+++ b/config.js
@@ -73,7 +73,7 @@ Config.prototype._seed = function _seed(seed) {
     seedOrDefault('gossipLogLevel', LoggingLevels.error);
     seedOrDefault('joinLogLevel', LoggingLevels.warn);
     seedOrDefault('membershipLogLevel', LoggingLevels.error);
-    seedOrDefault('ringLogLevel', LoggingLevels.debug);
+    seedOrDefault('ringLogLevel', LoggingLevels.error);
     seedOrDefault('suspicionLogLevel', LoggingLevels.error);
 
     // Gossip configs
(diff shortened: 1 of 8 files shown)


config.js
```diff
@@ -73,6 +73,7 @@ Config.prototype._seed = function _seed(seed) {
     seedOrDefault('gossipLogLevel', LoggingLevels.error);
     seedOrDefault('joinLogLevel', LoggingLevels.warn);
     seedOrDefault('membershipLogLevel', LoggingLevels.error);
+    seedOrDefault('ringLogLevel', LoggingLevels.debug);
```
thanodnl: Do we want this level enabled in all our users application?

lib/on_membership_event.js
```diff
@@ -23,9 +23,20 @@ var Member = require('./membership/member.js');
...
 
 function createChecksumComputedHandler(ringpop) {
-    return function onMembershipChecksumComputed() {
+    var membershipLogger = ringpop.loggerFactory.getLogger('membership');
+
+    return function onMembershipChecksumComputed(event) {
         ringpop.stat('increment', 'membership.checksum-computed');
         ringpop.emit('membershipChecksumComputed');
+
+        if (event.oldChecksum !== event.checksum) {
```
thanodnl: would it make sense to also `stat` an `increment` in here so we can graph the useless computations?

config.js
jwolski: No. Good point. I'm using this branch to double as a PR and as the ringpopd test branch. Thanks for calling this out. I'll change back to 'error' level before merging.

PR #229, opened 2015-12-02
Retry limit

jwolski:
We've been told that TChannel does a lot less bookkeeping if retry limit is set to 0. Doing so for Ringpop, especially it's gossip protocol, should/will reduce that amount of bookkeeping done.

@uber/ringpop


thanodnl:
lgtm

jwolski:
thanks for the quick "lgtm" @thanodnl

PR #230, opened 2015-12-02
validate source of joiner, entire cluster used to crash when sent an …

CorgiMan:
…invalid host-port as the source field

### DIFF
--- a/client.js
+++ b/client.js
@@ -20,6 +20,7 @@
 'use strict';
 
 var safeParse = require('./lib/util.js').safeParse;
+var validateHostPort = require('./lib/util.js').validateHostPort;
 var TChannel = require('tchannel');
 var TypedError = require('error/typed');
 
@@ -28,6 +29,12 @@ var ChannelDestroyedError = TypedError({
     message: 'Channel is already destroyed',
     endpoint: null,
     channelType: null
+});
+
+var InvalidHostPortError = TypedError({
+    type: 'ringpop.client.invalid-hostport',
+    message: 'Request made with invalid host port combination ({hostPort})',
+    hostPort: null
 });
 
 function RingpopClient(subChannel) {
@@ -121,7 +128,14 @@ RingpopClient.prototype._request = function _request(opts, endpoint, head, body,
         return;
     }
 
-    this.subChannel.waitForIdentified({
+    if (!opts || !validateHostPort(opts.host)) {
+        callback(InvalidHostPortError({
+            hostPort: String(opts && opts.host)
+        }));
+        return;
+    }
+
+    self.subChannel.waitForIdentified({
         host: opts.host
     }, function onIdentified(err) {
         if (err) {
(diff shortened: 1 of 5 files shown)


jwolski:
plz add a test

client.js
```diff
@@ -121,27 +121,36 @@ RingpopClient.prototype._request = function _request(opts, endpoint, head, body,
...
-        if (err) {
-            callback(err);
-            return;
-        }
+    try {
+        request();
+    } catch(e) {
+        callback(e);
+        return;
+    }
```
CorgiMan: Do not know if try catching every request is a good idea, we could also only try catch only pings and ping-reqs for example

Raynos: this is a bad idea. dont try catch...

CorgiMan: Hi Raynos, tchannel crashes when the hostport is malformed. In a rare case, where I tcurled with a malformed hostport field, I have crashed the entire cluster. This specific case doesn't happen anymore after we hardened the join mechanics of ringpop but I still want to strengthen this. Do you know another way?

Raynos: Crashes are bugs. File bug reports on tchannel.

Don't just swallow and hide bugs.

We continue on uncaught exceptions already so no need for this kind of defensive programming

jwolski: @Raynos TChannel asserts on its input and raises exceptions. Is there API documentation that specifies the pre-conditions of the TChannel API (particularly, `waitForIdentified`, `request`) Otherwise, if there's not, you can't expect users to trace through TChannel code with every release to make sure they are being as defensive as they need to be to make sure that TChannel doesn't crash their application.

Some apps _do not_ continue on uncaught.

There are the assertions that we are being "bit" by:

```
574     assert(parts.length === 2, 'invalid destination');
575     var host = parts[0];
576     var port = parts[1];
577     assert(host !== '0.0.0.0', 'cannot connect to ephemeral peer');
578     assert(port !== '0', 'cannot connect to dynamic port');
```

It seems a

server/protocol/join.js
```diff
@@ -43,13 +43,19 @@ var InvalidJoinAppError = TypedError({
     actual: null
 });
 
-var InvalidJoinSourceError = TypedError({
-    type: 'ringpop.invalid-join.source',
+var InvalidJoinSelfJoinError = TypedError({
```
jwolski: `SelfJoinError` is enough.

test/unit/index-test.js
```diff
@@ -475,3 +476,21 @@ test('first time member, not alive', function t(assert) {
     ringpop.destroy();
     assert.end();
 });
+
+test('don\'t call tchannel with invalid host port pair', function t(assert) {
```
jwolski: Test the case where no host value is provided aka `host: null`

client.js
Raynos: Applications that *do not* continue on uncaught is not something we should support. 

If you use ringpop; you continue on uncaught.

@jwolski we should definitely be more careful about host port assertions. we can document what a host:port is in the tchannel docs.

PR #234, opened 2015-12-15
Add a column for golang compatibility to the stats table.

thanodnl:


thanodnl:
Please ignore, I will move this PR to [`uber/ringpop-common`](https://github.com/uber/ringpop-common)

PR #236, opened 2016-01-19
aggresively raise piggyback counter when recipient of ping/ping-req

CorgiMan:


lib/gossip/dissemination.js
```diff
@@ -182,7 +194,7 @@ Dissemination.prototype._issueAs = function _issueAs(filterChange, mapChanges) {
             continue;
         }
 
-        if (change.piggybackCount > this.maxPiggybackCount) {
+        if (change.piggybackCount >= this.maxPiggybackCount) {
```
CorgiMan: We used to raise pbcounters before this statement, now we do it after.

jwolski:
unit test plz

lib/gossip/dissemination.js
thanodnl: What happens if we delete the item before we have confirmation that it has been processed by the other side? I know that that cannot be done with `issueAsReceiver`, but `_issueAs` is also used by `issueAsSender`. With this change the changes might be removed before we have confirmation back.

CorgiMan: I'm not sure I understand. The piggyback counter is only bumped when the communication is successful (with the exception on issueAsReceiver).
Previously the counter was  raised before this statement, in that case it was possible to throw the change away when we didn't have confirmation that the change was successfully disseminated.

CorgiMan: Previously the counter started at 1, because it was bumped before it hit this statement.
Now the counter is at 0 when it firstly hits this statement.

test/unit/dissemination-test.js
```diff
@@ -70,3 +70,80 @@ testRingpop('avoids redundant dissemination by filtering changes from source', f
...
+    dissemination.clearChanges();
+
+    membership.makeAlive(addrAlive, incNo);
+    membership.makeSuspect(addrSuspect, incNo);
+    membership.makeFaulty(addrFaulty, incNo);
+
+    // 'sender' and source of updates are different; issues changes.
+    var changes;
+
+    assert.equal(dissemination.maxPiggybackCount, 15, 'assert maxPiggybackCount is 15');
```
jwolski: I would suggest not referring to a magic number like `15` here. It leads to brittle assertions. I'd get rid of this assertion altogether or extract a method from `adjustMaxPiggybackCount` to compute the piggyback count based on known cluster size.

test/unit/dissemination-test.js
```diff
@@ -70,3 +70,80 @@ testRingpop('avoids redundant dissemination by filtering changes from source', f
...
+        onIssue(new Error('error so that piggyback counter isn\'t raised'));
+    });
+
+    var disChanges = dissemination.changes;
+    assert.equal(disChanges[addrAlive].piggybackCount, 0, 'piggyback counter isn\'t raised');
+    assert.equal(disChanges[addrSuspect].piggybackCount, 0, 'piggyback counter isn\'t raised');
+    assert.equal(disChanges[addrFaulty].piggybackCount, 0, 'piggyback counter isn\'t raised');
+
+    // Exhaust changes.
+    assert.equal(dissemination.maxPiggybackCount, 15, 'assert maxPiggybackCount is 15');
```
jwolski: Here too re: `15`.

test/unit/dissemination-test.js
```diff
@@ -70,3 +70,80 @@ testRingpop('avoids redundant dissemination by filtering changes from source', f
...
+    membership.makeSuspect(addrSuspect, incNo);
+    membership.makeFaulty(addrFaulty, incNo);
+
+    // 'sender' and source of updates are different; issues changes.
+    var changes;
+
+    assert.equal(dissemination.maxPiggybackCount, 15, 'assert maxPiggybackCount is 15');
+    for (var i = 0; i < dissemination.maxPiggybackCount;  i++) {
+        changes = dissemination.issueAsReceiver(addrAlive, incNo, membership.checksum);
+        assert.equal(changes.length, 3, 'changes issued');
```
jwolski: There's an implicit assumption being made here that the cluster / membership size is 3. To make this a bit more readable, can you do the following?

```
...
membership.makeAlive(...);
membership.makeSuspect(...);
membership.makeFaulty(...);
var expectedNumberOfChanges = membership.getMemberCount();

// then later...
assert.equal(changes.length, expectedNumberOfChanges);
```

Does that make sense?

jwolski:
Some minor comments about improving readability. Plz address.

test/unit/dissemination-test.js
CorgiMan: It does, thnx

test/unit/dissemination-test.js
CorgiMan: Will do

jwolski:
lgtm

test/unit/dissemination-test.js
```diff
@@ -70,3 +70,83 @@ testRingpop('avoids redundant dissemination by filtering changes from source', f
...
+        assert.equal(changes.length, expectedNumberOfChanges, 'expect ' + expectedNumberOfChanges + ' number of changes');
+        onIssue(new Error('error so that piggyback counter isn\'t raised'));
+    });
+
+    // 'sender' and source of updates are different; issues changes.
+    var disChangeAlive = dissemination.changes[addrAlive];
+    var disChangeSuspect = dissemination.changes[addrAlive];
+    var disChangeFaulty = dissemination.changes[addrFaulty];
+
+    assert.equal(disChangeAlive.piggybackCount, 0, "piggyback counter starts at 0");
```
thanodnl: nit: I only see strings with `'singel quotes'` in the rest of the file, and I think that is the preferred way here.

thanodnl:
lgtm

PR #238, opened 2016-01-29
Track potentially wedged requests and cancel them

jwolski:
This pull requests creates a trackable request object (`ClientRequest`). Inflight requests (that have not had their callback called) are placed into `requestsById` in the `Client` object. There is a background timer (`wedgedTimer`) that periodically scans for "wedged requests" and cancels the request by preemptively invoking its callback. A wedged requests is canceled if it has been inflight for greater than `wedgedRequestTimeout`.

Internally, `ClientRequest` maintains a "state machine" at each step of the TChannel-send process. When a `ClientRequest` is canceled, its callback will be invoked by `Client`. The `ClientRequest` also makes sure that a callback isn't called multiple times by nullifying the callback after its first invocation.

@uber/ringpop

### DIFF
--- a/client.js
+++ b/client.js
@@ -19,26 +19,26 @@
 // THE SOFTWARE.
 'use strict';
 
-var safeParse = require('./lib/util.js').safeParse;
-var validateHostPort = require('./lib/util.js').validateHostPort;
+var _ = require('underscore');
+var Config = require('./config.js');
+var ClientErrors = require('./client_errors.js');
+var ClientRequest = require('./client_request.js');
+var globalTimers = require('timers');
+var RingpopErrors = require('./ringpop_errors.js');
 var TChannel = require('tchannel');
-var TypedError = require('error/typed');
 
-var ChannelDestroyedError = TypedError({
-    type: 'ringpop.client.channel-destroyed',
-    message: 'Channel is already destroyed',
-    endpoint: null,
-    channelType: null
-});
-
-var InvalidHostPortError = TypedError({
-    type: 'ringpop.client.invalid-hostport',
-    message: 'Request made with invalid host port combination ({hostPort})',
-    hostPort: null
-});
-
-function RingpopClient(subChannel) {
+function RingpopClient(ringpop, subChannel, timers, date) {
+    this.ringpop = ringpop;
     this.subChannel = subChannel;
+    this.timers = timers || globalTimers;
+    this.date = date || Date;
+
+    this.config = this.ringpop.config;
+    this.logger = this.ringpop.loggerFactory.getLogger('client');
+
+    // If no subChannel provided, create one from a new instance
+    // of TChannel. This client then becomes the owner of that
+    // instance and is responsible for closing.
     this.isChannelOwner = false;
     if (!this.subChannel) {
         this.tchannel = new TChannel();
@@ -48,6 +48,10 @@ function RingpopClient(subChannel) {
         });
         this.isChannelOwner = true;
     }
+
+    this.requestsById = {};
+    this.wedgedTimer = null;
+    this.isDestroyed = false;
 }
 
 RingpopClient.prototype.adminConfigGet = function adminConfigGet(host, body, callback) {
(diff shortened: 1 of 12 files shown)


client_request_states.js
```diff
@@ -0,0 +1,28 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+'use strict';
+
+module.exports = {
+    initialized: 0,
+    waitForIdentifiedPre: 1,
+    waitForIdentifiedPost: 2,
+    subChannelRequestPre: 3,
+    subChannelRequestPost: 4 // This also represents TChannel completeness
```
benfleis: If you like, for clarity you can also define `completed: 4` -- not sure if it adds value/clarity or not.

client_request.js
```diff
@@ -0,0 +1,208 @@
...
+                endpoint: endpoint,
+                channelType: 'subChannel'
+            }));
+        });
+        return;
+    }
+
+    if (subChannel &&
+            subChannel.topChannel &&
+            subChannel.topChannel.destroyed) {
```
benfleis: question: can a topchannel be destroyed, without the subchannel having also been destroyed? ie, does the ordering here make it impossible for this 2nd one to ever fire?

client_request_states.js
jwolski: you mean in place of 'subChannelRequestPost` or in addition?

client_request.js
jwolski: That I am not sure of. I haven't revisited this logic in a while. The code was migrated (pretty much directly) from `client.js` into `client_request.js` in this PR. I'd rather not revisit the correctness of these checks now. I think the defensiveness is will be less harmful than good. But fair point to raise.

client.js
```diff
@@ -100,75 +112,94 @@ RingpopClient.prototype.destroy = function destroy(callback) {
...
 
-    if (this.subChannel &&
-            this.subChannel.topChannel &&
-            this.subChannel.topChannel.destroyed) {
+    var request = new ClientRequest(this, opts, endpoint, head, body, onSend);
+
+    // Evaluate if Ringpop should start applying backpressure on itself.
+    // Allow only a certain amount of potentially wedged requests to exist.
+    // It is highly unlikely that this'll ever be triggered, but it's best
+    // not to issue an unbounded number of wedged requests.
```
benfleis: this comment is slightly confusing to me -- we don't issue wedged requests, so much as issue requests and monitor for wedgies, right? :)

jwolski: yea "potentially wedged requests" is a bad way to put it. I'll fix.

client.js
```diff
@@ -100,75 +112,94 @@ RingpopClient.prototype.destroy = function destroy(callback) {
...
-            return;
+RingpopClient.prototype._scanForWedgedRequests = function _scanForWedgedRequests() {
+    var wedgedRequests = [];
+
+    // Find all requests that have not been timed-out by TChannel.
+    // Induce Ringpop timeout for wedged requests and cancel them.
+    var requestIds = Object.keys(this.requestsById);
+    for (var i = 0; i < requestIds.length; i++) {
+        var request = this.requestsById[requestIds[i]];
+        var timeInflight = Date.now() - request.timestamp;
```
benfleis: nit: should Date.now move outside of loop to `var now = Date.now();` and be used as such?

client.js
```diff
@@ -100,75 +112,94 @@ RingpopClient.prototype.destroy = function destroy(callback) {
...
-            callback(err);
-            return;
-        }
+    return wedgedRequests;
 
-        callback(null, safeParse(arg3));
+    // Log the first N requests only. We don't want to flood
+    // our pipes.
+    function wedgedRequestsToLog() {
+        return _.take(wedgedRequests, 10).map(function toLog(request) {
```
benfleis: If we log less than what we actually do, can we log the total number, so that we have some record of actual execution?

client.js
```diff
@@ -100,75 +112,94 @@ RingpopClient.prototype.destroy = function destroy(callback) {
...
     }
 };
 
+RingpopClient.prototype._scheduleWedgedTimer = function _scheduleWedgedTimer() {
+    var self = this;
+    this.wedgedTimer = this.timers.setTimeout(function onTimeout() {
+        if (self.isDestroyed) return;
+        self._scanForWedgedRequests();
+        self._scheduleWedgedTimer();
+    }, this.config.get('wedgedTimerInterval') || DEFAULT_WEDGED_TIMER_INTERVAL);
```
benfleis: suggest adding `this.wedgedTimer.unref();` afterward.

client_request_states.js
benfleis: in addition; i've seen protocol things like that that have an 'END' or 'COMPLETED' that matches the last state, so it's imho not an abnormal alias to add. it makes later condition checks read far more naturally, a la

```js
if (foo === State.completed) {
  ...
}
```

client_request.js
benfleis: yeah agreed. for style i would also think we would generally check the 'top level thing' first, but i am not suggesting that it change right now so much as raising it at the moment it occurred to me.

client.js
jwolski: good catch. will fix.

client.js
jwolski: done.

client.js
jwolski: :dancer:

client_request_states.js
```diff
@@ -0,0 +1,29 @@
...
+// THE SOFTWARE.
+'use strict';
+
+module.exports = {
+    initialized: 0,
+    waitForIdentifiedPre: 1,
+    waitForIdentifiedPost: 2,
+    subChannelRequestPre: 3,
+    subChannelRequestPost: 4, // This also represents TChannel completeness
+    completed: 4
```
thanodnl: The number 4 is in here twice, shouldn't this be 5?

config.js
```diff
@@ -119,6 +120,13 @@ Config.prototype._seed = function _seed(seed) {
     }, 'expected to be array of RegExp objects');
     seedOrDefault('membershipUpdateRollupEnabled', false);
     seedOrDefault('tchannelRetryLimit', 0);
+    // How long a request is outstanding until is it considered
+    // wedged and then canceled.
+    seedOrDefault('wedgedRequestTimeout', 30 * 1000);
```
thanodnl: does it make sense to reuse the constant from [`client.js`](https://github.com/uber/ringpop-node/pull/238/files#diff-215b25d5bf5c0b4623ad1a2b62456f60R36). Now we have `30s` defined twice.

client_request_states.js
jwolski: No. @benfleis had suggested that it's useful to know that the last actual step is equal to and synonymous with completion.

config.js
jwolski: yea, probably. I'll fix.

PR #240, opened 2016-02-03
Update README for newbies

Motiejus:
* Prepare ringpop for development.
* Update the max open file limit (needed in default settings of OSX).
* Introduce tick-cluster to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,32 @@ and provides request forwarding as a routing convenience. It can be used to
 shard your application in a way that's scalable and fault tolerant.
 
 # Installation
-To install Ringpop:
+To install Ringpop for usage as a library:
 
 ```
 npm install ringpop
+```
+
+Prepare the current directory for development:
+
+```
+npm install
+```
+
+To be able to run the tests, make sure you have your open file limit
+restriction on at least 4K:
+
+```
+ulimit -n 4096
+```
+
+# Tick Cluster
+An example application `scripts/tick-cluster.js` is included to this
+repository. It just launches a ringpop cluster of a given size. Using this
+application is the quickest way to start a ringpop cluster.
+
+```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -5,12 +5,34 @@ and provides request forwarding as a routing convenience. It can be used to
...
+ulimit -n 4096
+```
+
+# Tick Cluster
+An example application `scripts/tick-cluster.js` is included to this
+repository. It just launches a ringpop cluster of a given size. Using this
+application is the quickest way to start a ringpop cluster.
+
+```
+./scripts/tick-cluster.js --help
```
thanodnl: Can you add the command to start `tick-cluster` with the application in `main.js` it might not be as straight forward as one hopes.

Motiejus: Like discussed, I changed help line of `tick-cluster` to make it easier to understand what the interpreter is instead.

dansimau: +1 to Nils' original comment here. Don't make the reader think.

Motiejus: OK, let it be so. I uploaded a new revision.

dansimau:
LGTM :+1:

thanodnl:
lgtm

README.md
```diff
@@ -5,12 +5,34 @@ and provides request forwarding as a routing convenience. It can be used to
...
 
 # Installation
-To install Ringpop:
+To install Ringpop for usage as a library:
 
 ```
 npm install ringpop
 ```
 
+Prepare the current directory for development:
```
jwolski: I think this is a bit confusing. We should make clear distinction between Installation as a User and Setup for a Developer. The way the `npm install ringpop` and `npm install` instructions flow makes it seem like both are required.

I typically see these broken into 2 sections:

```
# Installation
```

and 

```
# Developing
```

PR #242, opened 2016-02-10
Add REPL over TCP support

Motiejus:


severb:
Do you mind adding a description of why this is useful? Is it for ninja debugging?

Motiejus:
Yep, it was meant to be for ninja debugging. But we had a chat with the team later, and agreed it should rather be part of ringpop-client. So closing.

PR #243, opened 2016-02-10
Update tchan to 3.6.13 and fix two tests that broke because of it.

CorgiMan:
One test used to add '001' to the hostport to make it unreachable, tchannel doesn't allow huge port no.s anymore. The other test failed because: 'AssertionError: TChannel must listen on top channel'. Listening to the rootChannel fixed the test.

### DIFF
--- a/test/lib/tchannel-proxy-cluster.js
+++ b/test/lib/tchannel-proxy-cluster.js
@@ -78,7 +78,7 @@ TChannelProxyCluster.prototype.bootstrap = function bootstrap(cb) {
         self.setupHandler('one', self.cluster.one);
         self.setupHandler('two', self.cluster.two);
         self.setupHandler('three', self.cluster.three);
-        self.client.listen(0, '127.0.0.1', onListen);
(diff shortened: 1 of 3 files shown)


CorgiMan:
@uber/ringpop

test/lib/tchannel-proxy-cluster.js
```diff
@@ -78,7 +78,7 @@ TChannelProxyCluster.prototype.bootstrap = function bootstrap(cb) {
         self.setupHandler('one', self.cluster.one);
         self.setupHandler('two', self.cluster.two);
         self.setupHandler('three', self.cluster.three);
-        self.client.listen(0, '127.0.0.1', onListen);
+        self.clientRootChannel.listen(0, '127.0.0.1', onListen);
```
CorgiMan: Is this ok? It fixes the test but I don't really know what I'm doing here...

jwolski: seems about right. @Raynos ??

jwolski:
lgtm

Raynos:
lgtm

PR #247, opened 2016-02-18
T287803 ring.checksum stats

benfleis:
add {membership,ring}-checksum{,-periodic} stats for partition detection

replaced time-mock with timer-shim, to work with intervals, added small wrapper for trivial compatibility.

@uber/ringpop

### DIFF
--- a/index.js
+++ b/index.js
@@ -172,6 +172,7 @@ function RingPop(options) {
     registerRingpopListeners(this);
 
     this.periodicStats = new PeriodicStats(this, {timers: timers});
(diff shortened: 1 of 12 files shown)


dansimau:
LGTM.

Can I suggest that we stop adding to `test/integration/` tests, and instead always add them to the shared integration tests? (@uber/ringpop)

Particularly because this feature has node/go parity, I feel there should be a shared integration test that both pass.

Motiejus:
lgtm

thanodnl:
lgtm

PR #248, opened 2016-03-17
Refactor the suspicion protocol to a generic timebased state transition.

severb:
This refactoring prepares the ground for reaping faulty nodes.


lib/gossip/state_transitions.js
```diff
@@ -60,33 +70,51 @@ Suspicion.prototype.start = function start(member) {
         return;
     }
 
-    if (this.timers[member.address]) {
+    var timer = this.timers[member.address];
+    if (timer) {
+        if (timer.state === state) {
+            this.logger.warn('redundant call to schedule a state transition for member, ignored', {
+                local: this.ringpop.whoami(),
+                suspect: member.address,
```
Motiejus: `suspect` here is very confusing. How about `address`?

lib/gossip/state_transitions.js
```diff
@@ -60,33 +70,51 @@ Suspicion.prototype.start = function start(member) {
...
+    this.timers[member.address] = {
+        timer: setTimeout(transition, timeout),
+        state: state
+    };
 
-    this.logger.debug('started suspect period', {
+    this.logger.debug('started state transition for member', {
         local: this.ringpop.whoami(),
-        suspect: member.address
+        suspect: member.address,
```
Motiejus: likewise, `member`

lib/gossip/state_transitions.js
```diff
@@ -60,33 +70,51 @@ Suspicion.prototype.start = function start(member) {
...
 };
 
-Suspicion.prototype.stop = function stop(member) {
-    clearTimeout(this.timers[member.address]);
+StateTransitions.prototype.stop = function stop(member) {
+    var timer = this.timers[member.address];
+    if (!timer) {
+        this.logger.warn('no state transition for member', {
+            local: this.ringpop.whoami(),
+            suspect: member.address
```
Motiejus: `member`

lib/gossip/state_transitions.js
severb: Good idea

test/unit/index-test.js
```diff
@@ -33,7 +33,7 @@ function createRingpop(opts) {
     var ringpop = new Ringpop(_.extend({
         app: 'test',
         hostPort: '127.0.0.1:3000'
-    }), opts);
+    }, opts));
```
Motiejus: Good catch.

Motiejus:
Besides the name in the log, LGTM.

lib/gossip/state_transitions.js
thanodnl: I think I renamed everything from `suspect` to `subject`.

lib/gossip/state_transitions.js
```diff
@@ -19,19 +19,20 @@
...
-    this.logger = this.ringpop.loggerFactory.getLogger('suspicion');
+    this.periods = options.periods || {};
+    this.logger = this.ringpop.loggerFactory.getLogger('statetransitions');
 
     this.isStoppedAll = null;
     this.timers = {};
 }
 
-Suspicion.prototype.reenable = function reenable() {
+StateTransitions.prototype.reenable = function reenable() {
```
thanodnl: There was a whole discussion in the ringpop-go repo about naming, one of the outcomes of that was to name this `enable`.

lib/gossip/state_transitions.js
severb: In here everything was called `member`, and I left it as it was. Do you think we should be consistent in both codebases and rename `member` to `subject`?

lib/gossip/state_transitions.js
```diff
@@ -60,33 +70,51 @@ Suspicion.prototype.start = function start(member) {
...
         local: this.ringpop.whoami(),
-        suspect: member.address
+        suspect: member.address,
+        state: state
     });
 };
 
-Suspicion.prototype.stop = function stop(member) {
-    clearTimeout(this.timers[member.address]);
+StateTransitions.prototype.stop = function stop(member) {
```
thanodnl: same naming discussion; `cancel` over `stop`

lib/gossip/state_transitions.js
severb: I like that, I'll do the same here.

severb:
@Motiejus @thanodnl can you take a second look?

lib/gossip/state_transitions.js
```diff
@@ -0,0 +1,144 @@
...
+    this.periods = options.periods || {};
+    this.logger = this.ringpop.loggerFactory.getLogger('statetransitions');
+
+    this.isStoppedAll = null;
+    this.timers = {};
+}
+
+StateTransitions.prototype.enable = function enable() {
+    if (this.isStoppedAll !== true) {
+        this.logger.debug('ringpop cannot enable state transitions because it was never disabled', {
```
Motiejus: This seems like invariant violation -- maybe `warning` or `error` should be more appropriate.

lib/gossip/state_transitions.js
```diff
@@ -0,0 +1,144 @@
...
+    var timeout = this.periods.suspect || StateTransitions.Defaults.suspect;
+    this.schedule(member, Member.Status.suspect, timeout, function() {
+        self.ringpop.membership.makeFaulty(member.address,
+            member.incarnationNumber);
+    });
+};
+
+StateTransitions.prototype.schedule = function schedule(member, state, timeout, transition) {
+    if (this.isStoppedAll === true) {
+        this.logger.debug('cannot start a state transition because state transitions has not been enabled', {
```
Motiejus: same here, maybe `warning` or `error` instead, since we don't expect this to happen normally?

lib/gossip/state_transitions.js
```diff
@@ -0,0 +1,144 @@
...
+'use strict';
+
+var Member = require('../membership/member');
+
+function StateTransitions(options) {
+    this.ringpop = options.ringpop;
+    this.periods = options.periods || {};
+    this.logger = this.ringpop.loggerFactory.getLogger('statetransitions');
+
+    this.isStoppedAll = null;
```
thanodnl: can we call this variable `enabled` now that the functions to control them are named `enable` and `disable`. This seems to be the only occurrence regarding `stopped`.

lib/gossip/state_transitions.js
```diff
@@ -0,0 +1,144 @@
...
+
+    this.logger.debug('enabled state transitions protocol', {
+        local: this.ringpop.whoami()
+    });
+};
+
+StateTransitions.prototype.scheduleSuspectToFaulty = function scheduleSuspectToFaulty(member) {
+    var self = this;
+    var timeout = this.periods.suspect || StateTransitions.Defaults.suspect;
```
thanodnl: is it expected that the `periods` are changed after creation? Otherwise I would suggest to merge these values in the constructor already. It makes the code easier to follow.

test/unit/index-test.js
```diff
@@ -155,7 +155,7 @@ test('admin leave stops gossip', function t(assert) {
     });
 });
 
-test('admin leave stops suspicion subprotocol', function t(assert) {
+test('admin leave stops state tranisitons', function t(assert) {
```
Motiejus: typo

Motiejus:
See some non-blocking comments, so LGTM.

For next time, it would be great to have a non-squashed commit for re-review (since seems like github doesn't support looking between two edits of the same commit).

index.js
```diff
@@ -42,7 +42,7 @@ var metrics = require('metrics');
 var packageJSON = require('./package.json');
 
 var Gossip = require('./lib/gossip');
-var Suspicion = require('./lib/gossip/suspicion');
+var StateTransitions = require('./lib/gossip/state_transitions');
```
thanodnl: Not really blocking because it is used both in this project, but normally nodejs modules name files with `-` in favor of `_`. In ringpop they are both used which is already quite messy.

lib/gossip/state_transitions.js
```diff
@@ -0,0 +1,146 @@
...
+StateTransitions.prototype.scheduleSuspectToFaulty = function scheduleSuspectToFaulty(member) {
+    var self = this;
+    this.schedule(member, Member.Status.suspect, this.suspectTimeout, function() {
+        self.ringpop.membership.makeFaulty(member.address,
+            member.incarnationNumber);
+    });
+};
+
+StateTransitions.prototype.schedule = function schedule(member, state, timeout, transition) {
+    if (this.enabled === true) {
```
thanodnl: This prevents scheduling when the controller is enabled! Please add appropriate tests if they currently pass.

lib/gossip/state_transitions.js
```diff
@@ -0,0 +1,146 @@
...
+
+function StateTransitions(options) {
+    this.ringpop = options.ringpop;
+
+    var p = options.periods || {};
+    this.suspectTimeout = p.suspect || StateTransitions.Defaults.suspect;
+
+    this.logger = this.ringpop.loggerFactory.getLogger('statetransitions');
+
+    this.enabled = null;
```
thanodnl: can we use `boolean` types here. Even though node does not really care it is easier to understand.

lib/gossip/state_transitions.js
```diff
@@ -0,0 +1,146 @@
...
+    this.timers = {};
+}
+
+StateTransitions.prototype.enable = function enable() {
+    if (this.enabled !== true) {
```
thanodnl: `enabled` seems to be used throughout the code in a way that does not make sense for the name of the application. where `enabled` is set to `true` to **disable** the module.

severb: Yeah, this is confusing, I'll change it. It was this way because of a previous rename that was negating the current name.

thanodnl:
LGTM

severb:
Ty @Motiejus @thanodnl

PR #250, opened 2016-04-11
Add server/client middleware to patch tombstone state

severb:
This change introduces a middleware to both the ringpop server and tchannel client.

A middleware is an object that implements *any* of the following methods:
* `request(req, arg2, arg3, callback)`
* `response(req, err, res1, res2, callback)`

A middleware is free to alter `req` and update/replace any of `arg2`, `arg3`, `err`, `res1` or `res2`. Multiple middlewares can be chained together forming a middleware stack. The `callback` is used to pass the control to the next middleware in chain. For example, a noop middleware can look like this:

```javascript
var noop = {
  'request': function(req, arg2, arg3, callback) {
    callback(arg2, arg3);
  },
  'response': function(req, err, res1, res2, callback) {
    callback(err, res1, res2);
  }
};
```

Given the middleware stack defined as `MiddlewareStack([m1, m2]).run(req, arg2, arg3, handler, handler_callback)`, the order in which the middlewares are applied in is: `m1.request -> m2.request -> handler -> m2.response -> m1.response -> handler_callback`.

The difference between applying a middleware stack to the server and the client is the logical direction a request and a response point to. For example, for a server it looks like this: `server <- request`, `server -> response`; while for a client it's the reverse: `client -> request`, `client <- response`.

The middleware adds a convenient place to do global operations right at the ringpop boundaries in a way that's transparent and composable for the rest of the library. For example, the tombstone wire compatibility change is implemented in a middleware, allowing the rest of the code to be written as if the tombstone is a first class state like the others. Later, when we decide to gossip it as such, in a backward incompatible change, the only change needed is to remove the patching middleware.

The current implementation has a couple of shortcomings that I plan to address if the general feedback is positive:
* The middleware chaining should use `nextTick` to avoid blowing the call stack
* ~~Because of how the rest of the code is written, server middlewares must decode/encode any arguments or responses every time, which is inefficient. Ideally, we can push this processing in a different middleware that does decoding/encoding as soon/late as possible.~~ ✅ 

In future, if the idea proves to be useful for other things, maybe we can push a similar functionality upstream in tchannel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "check-licence": "uber-licence --dry",
     "cover": "istanbul cover --print detail --report html test/index.js | faucet",
     "jshint": "jshint --verbose *.js lib/**/*.js scripts/*.js server/**/*.js",
-    "travis": "npm run cover -s && istanbul report lcov && ((cat coverage/lcov.info | coveralls) || exit 0)",
+    "travis": "npm run jshint && npm run cover -s && istanbul report lcov && ((cat coverage/lcov.info | coveralls) || exit 0)",
     "view-cover": "opn coverage/index.html"
   },
   "dependencies": {
@@ -54,7 +54,7 @@
     "glob": "^4.3.1",
     "istanbul": "^0.3.5",
     "itape": "^1.5.0",
-    "jshint": "^2.5.6",
+    "jshint": "^2.9.2",
     "leaked-handles": "^5.1.0",
     "opn": "^1.0.1",
     "pre-commit": "^0.0.9",


server/middleware.js
```diff
@@ -0,0 +1,163 @@
...
+                callResponseMiddleware(err, res1, res2);
+            }
+        } else {
+            callback(req, err, res1, res2)
+        }
+    }
+}
+
+
+function faultyAsTombstone(change) {
```
Motiejus: I think this should go to a separate file, outside of the "middleware core" (i.e. `thombstoning.js`).

dansimau:
I like the middleware approach. However, as you mentioned, I am concerned with a possible performance regression. Can we write some benchmarks to compare and measure performance overhead?

severb:
@dansimau I think it shouldn't be too hard to write some sort of performance test for this change. Do we already have anything in place to help with this sort of measurement?

Motiejus:
Disclaimer: I am a big fan of middlewares in these kinds of applications.

If we introduce middlewares now in the current form, they will feel like out of place, like you correctly identified :) decoding/encoding is done on every middleware, which is a sign that they are not first-class citizens (otherwise we would have encoding/decoding middleware, and can plug our business logic before or after it). If we want to adopt middlewares "proper", we should refactor that.

Also, in the current form, short-circuiting middleware is not possible (i.e. stop request processing). This would be useful if, e.g., we want to implement transparent auth. I think you could at least document semantics on how short-circuiting could look like from the library perspective.

severb:
@Motiejus Initially, I had a short-circuit mechanism in place but I thought it will just add more complexity with no clear benefit for my specific case. It worked by passing an additional callback to both `request` and `response`. I can easily add it back for the sake of completeness if deemed worthy.

Before that, I think we should agree if it's worth keeping the middleware as it is, ditch it and do the tombstone patching everywhere in the code where necessary, or improve by refactoring the code to at least do the encoding/decoding in a middleware so we don't have to pay the price multiple times.

From pure pragmatic POV, I think keeping it as it is, is still a step forward without paying the refactoring price. We can always improve/extract other middlewares in the future. As @dansimau suggested, I'll do a benchmark to see how high the price of multiple encoding/decoding is and if we can live with it.

severb:
I moved JSON encoding/decoding in a middleware. This solves the double encode/decode problem which was a potential source of performance regression. We still have to decide if we want the short-circuit mechanism in place at this time and fix the tests that stops me from using `nextTick` in the middleware stack. @Motiejus, @dansimau can you give it another look?

server/protocol/join.js
```diff
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 'use strict';
 
-var safeParse = require('../../lib/util').safeParse;
+//var safeParse = require('../../lib/util').safeParse;
```
Motiejus: Remove?

Motiejus:
@severb overall looks great, a few comments:

* how about moving each middleware (`thombstone_middleware.js` ,`req3-recoding.js`) to their own files?
* can you add some documentation on middlewares, and how to use them? `middlewares.js` is a good candidate for it, as I don't see developer-oriented documentation for ringpop-node.
* you are right about short-circuiting, we don't have to add it now.
* did you consider adding tests for `transportServerMiddleware`?

severb:
@Motiejus thanks,
* I think in the end there will be a middleware directory and each implementation in its own file
* The documentation is lacking, I'll add some comments to explain how things work
* Good call re. tests for `transportServerMiddleware`, I'll add some

client_request.js
```diff
@@ -27,9 +27,10 @@ var safeParse = require('./lib/util.js').safeParse;
 var validateHostPort = require('./lib/util.js').validateHostPort;
 var uuid = require('node-uuid');
 var util = require('util');
+var middleware = require('./lib/middleware');
 
-/* jshint maxparams: 7 */
-function ClientRequest(client, opts, endpoint, head, body, date, callback) {
+/* jshint maxparams: 8 */
+function ClientRequest(client, opts, endpoint, head, body, date, callback, middlewares) {
```
thanodnl: 'normal' node behavior is to always have the callback as the last function. If you want to make `middlewares` optional you test if `middlewares` is a function and in that case assign middlewares to the callback and then put a default value in the optional field.

client.js
```diff
@@ -27,7 +27,7 @@ var globalTimers = require('timers');
 var RingpopErrors = require('./ringpop_errors.js');
 var TChannel = require('tchannel');
 
-/*jshint maxparams: 5 */
+/* jshint maxparams: 8 */
```
Motiejus: Why did this change? I see only 5 params in this function.

lib/middleware/index.js
```diff
@@ -0,0 +1,106 @@
...
+ * var noop = {
+ *     'request': function(req, arg2, arg3, callback) {
+ *         callback(arg2, arg3);
+ *     },
+ *     'response': function(req, err, res1, res2, callback) {
+ *         callback(err, res1, res2);
+ *     }
+ * };
+ *
+ * Given the middleware stack defined as:
```
Motiejus: It would be useful to define `m1` and `m2` before this example. It is clear that these are two middlewares when you understand it, but it's helpful to know what they are _in order to_ understand them.

lib/middleware/index.js
```diff
@@ -0,0 +1,106 @@
...
+ * client is the logical direction a request and a response point to. For
+ * example, for a server it looks like this: server<-request, server->response;
+ * while for a client it's the reverse: client->request, client<-response.
+ */
+
+function MiddlewareStack(middlewares) {
+    this.middlewares = middlewares || [];
+}
+
+// XXX: adding nextTick to avoid blowing the call stack breaks a test
```
Motiejus: Which test?

lib/middleware/index.js
```diff
@@ -0,0 +1,106 @@
...
+            callback(req, err, res1, res2);
+        }
+    }
+};
+
+
+module.exports = _.extend({
+        'MiddlewareStack': MiddlewareStack
+    },
+    require('./tombstone.js'),
```
Motiejus: Do we need to include the actual middleware implementations here? Can we make this a "pure drop-in library"?

Maybe `index.js` could import all three modules, and this file (w/o imports) could become `core.js`?

Motiejus:
Apart the naming comment, LGTM.

client.js
severb: Good catch, I think this is a mistake.

client_request.js
```diff
@@ -112,49 +114,55 @@ ClientRequest.prototype.send = function send() {
...
-            host: opts.host,
-            serviceName: 'ringpop',
-            hasNoParent: true,
-            retryLimit: opts.retryLimit || 0,
-            trace: false,
-            headers: {
-                as: 'raw',
-                cn: 'ringpop'
+        self.middlewareStack.run(self, self.head, self.body,
+            function(req, arg2, arg3, callback) {
```
CorgiMan: should we name this function for debugging purposes?

client_request.js
```diff
@@ -112,49 +114,55 @@ ClientRequest.prototype.send = function send() {
...
-
-    function onSend(err, res, arg2, arg3) {
-        self._changeState(ClientRequestStates.subChannelRequestPost);
-
-        if (self.isCanceled) {
-            self.client.logger.error('ringpop tchannel completed after request was canceled', {
-                local: self.client.ringpop.whoami(),
-                state: self.state,
-                request: self.toLog()
+            function(req, err, res1) { // we don't use res2
```
CorgiMan: should we name this function for debugging purposes?

server/protocol/ping.js
```diff
@@ -20,7 +20,7 @@
 'use strict';
 
 var errors = require('../../lib/errors.js');
-var safeParse = require('../../lib/util').safeParse;
+//var safeParse = require('../../lib/util').safeParse;
```
CorgiMan: remove please

lib/middleware/index.js
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// XXX: adding nextTick to avoid blowing the call stack breaks a test
+MiddlewareStack.prototype.run = function run(req, arg2, arg3, handler, callback) {
+    var self = this;
+
+    var i = -1;
+    callRequestMiddleware(arg2, arg3);
+
+    function callRequestMiddleware(arg2, arg3) {
```
CorgiMan: I found the recursion here to be hard to process. I think the async lib could clear it up. However, if I'm fine with it like this as well if you add a little comment on how this recursion works.

lib/middleware/tombstone.js
```diff
@@ -0,0 +1,98 @@
...
+// THE SOFTWARE.
+'use strict';
+/* jshint maxparams: 5 */
+
+
+var Member = require('../membership/member.js');
+
+
+function faultyAsTombstone(change) {
+    if(change.status === Member.Status.faulty && change.tombstone) {
```
CorgiMan: s/if(/if (/
(add space please)

lib/middleware/tombstone.js
```diff
@@ -0,0 +1,98 @@
...
+        change.status = Member.Status.tombstone;
+        delete change.tombstone;
+        return true;
+    }
+    return false;
+}
+
+function tombstoneAsFaulty(change) {
+    if(change.status === Member.Status.tombstone) {
```
CorgiMan: s/if(/if (/
(add space please)

CorgiMan:
Wow, nice job! I was skeptical at first (as a good physicist should be) but this is awesome!

Looks good after you have addressed the feedback.

I think we really have to put something in the schema's of the integration tests that fails when it sees a tombstone field. Is that already in the pipeline?

severb:
@CorgiMan Thanks for the review! The tombstone part is largely untested (at least not e2e) as it is, but as soon as the next PR that implements tombstones lands, we have quite a few integration tests that exercise this part of the code as well; and they all pass.

lib/middleware/index.js
severb: Totally agree. I'm not familiar with the async lib, but thanks for the tip. I'll take a look and see if it can improve the readability. Either way, I think a comment is in place here.

PR #253, opened 2016-04-22
Use a DiscoverProvider-function to discover the bootstrap hosts

mennopruijssers:
By introducing the concept of DiscoverProviders in ringpop-node, we make it possible in the future to reload the host list. This is a requirement for partition healing.

When the number of hosts or their ip/ports are more dynamic (e.g. when bootstrapping over hyperbahn) it's going to be useful to reload the hosts after bootstrapping.


index.js
```diff
@@ -500,45 +517,24 @@ RingPop.prototype.lookupN = function lookupN(key, n) {
 };
 
 RingPop.prototype.reload = function reload(file, callback) {
```
mennopruijssers: I'm not sure if/how we want to support this function. It's currently called from the `admin/reload`-endpoint but only supports a file.

discover-providers.js
```diff
@@ -0,0 +1,120 @@
...
+        return callback(null, hosts);
+    };
+}
+
+/**
+ * Creates a DiscoverProvider from the provided options.
+ *
+ * @param opts: an object configuring the DiscoverProvider:
+ *          - if opts or opts.discoverProvider is a function, it's returned as the {DiscoverProvider}.
+ *          - if opts.hosts is an array, a static hosts discover provider is returned (see {createStaticHostsProvider})
```
thanodnl: if opts or opts.hosts ....

discover-providers.js
```diff
@@ -0,0 +1,120 @@
...
+ *          - if opts.bootstrapFile is a string, a json file discover provider is returned (see {createJsonFileDiscoverProvider})
```
thanodnl: if opts or opts.bootstrapFile ....

index.js
thanodnl: I think the right thing to do here is to add a `// Deprecated` comment to the `/admin/reload` handler and add a warning log when that endpoint is called. Later we can remove it.

thanodnl:
LGTM

discover-providers.js
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    return null;
+}
+
+module.exports = {
+    createFromOpts: createFromOpts,
+    createStaticHostsProvider: createStaticHostsProvider,
+    createJsonFileDiscoverProvider: createJsonFileDiscoverProvider
+};
```
dansimau: Missing EOL

index.js
dansimau: Agree with @thanodnl. I'd also check with @jwolski that it's OK to deprecate. I.e. how confident are we this is not in use?

test/unit/discover-providers-example.json
```diff
@@ -0,0 +1 @@
+["127.0.0.1:3000", "127.0.0.1:3001"]
```
dansimau: Missing EOL

index.js
mennopruijssers: It's not used in ringpop-admin and (according to our stats backend) never called in production.

mennopruijssers:
Based on feedback from @jwolski, I moved the deprecation of the endpoint to #255

mennopruijssers:
@dansimau All your (style) feedback is taking care of! Let's try to come up with a better linter ;).

dansimau:
LGTM 👍

PR #255, opened 2016-04-26
Deprecate "admin/reload"-endpoint

mennopruijssers:
Part of the feedback on #253 was to deprecate the `admin/reload`-endpoint.
The endpoint isn't used by ringpop-admin and, according to our internal metrics, has never been called in production.
Triggering a reload was probably useful during testing when the bootstrap file changed. With partition healing, we achieve a similar result in that the discover provider will be re-queried, except that it's automatic.

### DIFF
--- a/server/admin/reload.js
+++ b/server/admin/reload.js
@@ -24,6 +24,7 @@ var safeParse = require('../../lib/util.js').safeParse;
 
 module.exports = function createReloadHandler(ringpop) {
     return function handleReload(arg1, arg2, hostInfo, callback) {
+        ringpop.logger.warn('endpoint is deprecated: /admin/reload', {hostInfo: hostInfo});
         var body = safeParse(arg2.toString());
         if (body && body.file) {
             ringpop.reload(body.file, function(err) {
--- a/server/admin/reload.js
+++ b/server/admin/reload.js
@@ -24,7 +24,10 @@ var safeParse = require('../../lib/util.js').safeParse;
 
 module.exports = function createReloadHandler(ringpop) {
     return function handleReload(arg1, arg2, hostInfo, callback) {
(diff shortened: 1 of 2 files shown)


dansimau:
+1 for separate PR.

I'd like to add one more justification for deprecating this: triggering a reload was probably useful during testing when the bootstrap file changed. With partition healing, we achieve a similar result in that the discover provider will be re-queried, except that it's automatic.

If we deprecate this, perhaps we should remove these references too:
* https://github.com/uber/ringpop-common/blob/master/test/admin-tests.js#L31
* https://github.com/uber/ringpop-common/blob/master/test/ringpop-assert.js#L743

LGTM

mennopruijssers:
Good catch on the references in ringpop-common! I opened a PR to remove them (uber/ringpop-common#49)!

mennopruijssers:
and changed the description to include your justification :) (I'll make sure it makes the commit message as well ;))

server/admin/reload.js
```diff
@@ -24,6 +24,7 @@ var safeParse = require('../../lib/util.js').safeParse;
 
 module.exports = function createReloadHandler(ringpop) {
     return function handleReload(arg1, arg2, hostInfo, callback) {
+        ringpop.logger.warn('endpoint is deprecated: /admin/reload', {hostInfo: hostInfo});
```
jwolski: I would suggest changing the logging message to:

```
ringpop.logger.warn('ringpop endpoint is deprecated: /admin/reload', {
    local: ringpop.whoami(),
    hostInfo: hostInfo
});
```

The differences are including "ringpop" in the message itself. Since Ringpop is embedded in apps and its loggers are shared, it's typically not apparent whether the app or Ringpop is logging the message. The other difference is to include the `local` parameter as a logger metadata option because it's nice to know which Ringpop node is logging the message.

mennopruijssers: Makes sense. Thanks!

PR #257, opened 2016-05-02
Implement reaping faulty nodes

severb:
This adds the reaping faulty nodes on top of the tombstone wire compatibility (which is landed). It's very similar to the [Go implementation](https://github.com/uber/ringpop-go/pull/132) and includes the admin endpoint.

### DIFF
--- a/lib/gossip/state_transitions.js
+++ b/lib/gossip/state_transitions.js
@@ -27,6 +27,8 @@ function StateTransitions(options) {
 
     var p = options.periods || {};
     this.suspectTimeout = p.suspect || StateTransitions.Defaults.suspect;
+    this.faultyTimeout = p.faulty || StateTransitions.Defaults.faulty;
+    this.tombstoneTimeout = p.tombstone || StateTransitions.Defaults.tombstone;
 
     this.logger = this.ringpop.loggerFactory.getLogger('statetransitions');
 
@@ -54,6 +56,21 @@ StateTransitions.prototype.scheduleSuspectToFaulty = function scheduleSuspectToF
     this.schedule(member, Member.Status.suspect, this.suspectTimeout, function() {
         self.ringpop.membership.makeFaulty(member.address,
             member.incarnationNumber);
+    });
+};
+
+StateTransitions.prototype.scheduleFaultyToTombstone = function scheduleFaultyToTombstone(member) {
+    var self = this;
+    this.schedule(member, Member.Status.faulty, this.faultyTimeout, function() {
+        self.ringpop.membership.makeTombstone(member.address,
+            member.incarnationNumber);
+    });
+};
+
+StateTransitions.prototype.scheduleTombstoneToEvict = function scheduleTombstoneToEvict(member) {
+    var self = this;
+    this.schedule(member, Member.Status.tombstone, this.tombstoneTimeout, function() {
+        self.ringpop.membership.evict(member.address);
     });
 };
 
@@ -141,7 +158,9 @@ StateTransitions.prototype.disable = function disable() {
 };
 
 StateTransitions.Defaults = {
-    suspect: 5000
+    suspect: 5 * 1000, //5s
+    faulty: 24 * 60 * 60 * 1000, //24h
+    tombstone: 60 * 1000, //60s
 };
 
 module.exports = StateTransitions;
--- a/lib/gossip/state_transitions.js
+++ b/lib/gossip/state_transitions.js
@@ -21,14 +21,15 @@
 'use strict';
 
 var Member = require('../membership/member');
+var _ = require('underscore');
 
 function StateTransitions(options) {
     this.ringpop = options.ringpop;
 
-    var p = options.periods || {};
-    this.suspectTimeout = p.suspect || StateTransitions.Defaults.suspect;
-    this.faultyTimeout = p.faulty || StateTransitions.Defaults.faulty;
-    this.tombstoneTimeout = p.tombstone || StateTransitions.Defaults.tombstone;
+    var p = _.defaults({}, options.periods, StateTransitions.Defaults);
+    this.suspectTimeout = p.suspect;
+    this.faultyTimeout = p.faulty;
+    this.tombstoneTimeout = p.tombstone;
 
     this.logger = this.ringpop.loggerFactory.getLogger('statetransitions');
 
@@ -53,7 +54,7 @@ StateTransitions.prototype.enable = function enable() {
 
 StateTransitions.prototype.scheduleSuspectToFaulty = function scheduleSuspectToFaulty(member) {
     var self = this;
-    this.schedule(member, Member.Status.suspect, this.suspectTimeout, function() {
+    this.schedule(member, Member.Status.suspect, this.suspectTimeout, function suspectToFaulty() {
         self.ringpop.membership.makeFaulty(member.address,
             member.incarnationNumber);
     });
@@ -61,7 +62,7 @@ StateTransitions.prototype.scheduleSuspectToFaulty = function scheduleSuspectToF
 
 StateTransitions.prototype.scheduleFaultyToTombstone = function scheduleFaultyToTombstone(member) {
     var self = this;
-    this.schedule(member, Member.Status.faulty, this.faultyTimeout, function() {
+    this.schedule(member, Member.Status.faulty, this.faultyTimeout, function faultyToTombstone() {
         self.ringpop.membership.makeTombstone(member.address,
             member.incarnationNumber);
     });
@@ -69,7 +70,7 @@ StateTransitions.prototype.scheduleFaultyToTombstone = function scheduleFaultyTo
 
 StateTransitions.prototype.scheduleTombstoneToEvict = function scheduleTombstoneToEvict(member) {
     var self = this;
-    this.schedule(member, Member.Status.tombstone, this.tombstoneTimeout, function() {
+    this.schedule(member, Member.Status.tombstone, this.tombstoneTimeout, function tombstoneToEvict() {
         self.ringpop.membership.evict(member.address);
     });
 };
@@ -158,9 +159,9 @@ StateTransitions.prototype.disable = function disable() {
 };
 
 StateTransitions.Defaults = {
-    suspect: 5 * 1000, //5s
-    faulty: 24 * 60 * 60 * 1000, //24h
-    tombstone: 60 * 1000, //60s
+    suspect: 5 * 1000, // 5s
+    faulty: 24 * 60 * 60 * 1000, // 24h
+    tombstone: 60 * 1000, // 60s
 };
 
 module.exports = StateTransitions;
--- a/lib/membership/index.js
+++ b/lib/membership/index.js
@@ -132,6 +132,12 @@ Membership.prototype.generateChecksumString = function generateChecksumString() 
     for (var i = 0; i < sortedMembers.length; ++i) {
         var member = sortedMembers[i];
 
+        // Don't include Tombstone nodes in the checksum to avoid
+        // bringing them back to life through full syncs
+        if (member.status === Member.Status.tombstone) {
+            continue;
+        }
+
         checksumString += member.address +
             member.status +
             member.incarnationNumber + ';';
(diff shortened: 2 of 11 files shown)


lib/membership/index.js
```diff
@@ -230,6 +236,31 @@ Membership.prototype.makeSuspect = function makeSuspect(address, incarnationNumb
...
+    this.members = _.without(this.members, memberToDelete);
+    this.computeChecksum();
+};
+
+Membership.prototype.evict = function evict(address) {
+    if (this.localMember.address === address) {
+        this.logger.error('ringpop tried to evict the local member from the memberlist, action has been prevented');
+        return;
+    }
+    this.removeMember(address);
```
Motiejus: Why do we need `removeMember`? Can we just move the code here?

Reason: when I saw both functions, I started wondering "why are they different?", only to find out that one of them is a helper, not used by anything else.

lib/membership/member.js
```diff
@@ -173,7 +173,7 @@ Member.prototype._applyUpdatePenalty = function _applyUpdatePenalty() {
 Member.prototype._isLocalOverride = function _isLocalOverride(update) {
```
Motiejus: I was having a hard time understanding this function, would you mind adding this comment? https://github.com/uber/ringpop-node/pull/261

lib/on_membership_event.js
```diff
@@ -137,10 +142,12 @@ function createUpdatedHandlerForRing(ringpop) {
             var update = updates[i];
             switch (update.status) {
                 case Member.Status.alive:
+                // TODO: Faulty should be here
```
Motiejus: Can you elaborate? Why is this TODO here?

main.js
```diff
@@ -49,7 +49,11 @@ function main(args) {
             serviceName: 'ringpop',
             trace: false
         }),
-        isCrossPlatform: true
+        isCrossPlatform: true,
+        stateTimeouts: {
+            faulty: 5 * 1000, //5s
```
Motiejus: please add spaces after `//`

server/admin/member.js
```diff
@@ -107,5 +128,9 @@ module.exports = {
     memberLeave: {
         endpoint: '/admin/member/leave',
         handler: createLeaveHandler
+    },
+    reap: {
+        endpoint: '/admin/reap',
+        handler: createReapHandler
```
Motiejus: If the user calls `reap` on non-fauty node, the callback should return an error instead of OK.

test/unit/gossip_test.js
```diff
@@ -197,3 +197,47 @@ testRingpop({
...
+    var member = membership.findMemberByAddress(address);
+
+    stateTransitions.tombstoneTimeout = 1;
+    stateTransitions.scheduleTombstoneToEvict(member);
+
+    setTimeout(function onTimeout() {
+        assert.notOk(membership.findMemberByAddress(address), 'member was evicted');
+        done();
+    }, stateTransitions.period + 1);
+});
```
Motiejus: How long do these tests execute? I don't see you overwriting this value anywhere in the test setup.

server/admin/member.js
Motiejus: Also add a test for `/admin/reap`, with 3 cases:

1. trying to reap self.
2. trying to reap non-faulty.
3. trying to reap faulty.

lib/on_membership_event.js
severb: I think I might just go ahead and add it, I think @thanodnl added it in ringpop-go; it's also the correct behaviour.

server/admin/member.js
severb: I think we have a few integration tests for the admin endpoint.

test/unit/gossip_test.js
severb: There is no significant change in the suite run time because the test is async. But this is a good catch, we should use `stateTransitions.tombstoneTimeout` instead of `stateTransitions.period` (which is undefined).

There is another test where this we use `stateTransitions.period` instead of the right timer (the one I copied as a template for this test). Please take a look at this PR (#262) and give it a thumbs up so we fix it there too.

lib/membership/member.js
mennopruijssers: @Motiejus the comment in #261 on `_isLocalOverride` is not. If `_isLocalOverride` returns true, the update will be changed to assert aliveness (and bump incarnationNumber); if not and it's also not an other overwrite, it's skipped.

lib/membership/index.js
severb: I have no prob. with moving it. I just emulated the Go implementation; as for why I originally made them separate in Go, I can't remember. Maybe I thought I'm going to reuse it somewhere.

server/admin/member.js
severb: I don't understand the first comment. You can't call `/admin/reap` on faulty nodes. What this endpoint does, it speeds up the transition from faulty to tombstone (which by default takes 24h) for the node you're contacting. As soon as this happens, it will disseminate this to the entire cluster, causing everyone to reap their faulty nodes.

lib/gossip/state_transitions.js
```diff
@@ -57,6 +59,21 @@ StateTransitions.prototype.scheduleSuspectToFaulty = function scheduleSuspectToF
     });
 };
 
+StateTransitions.prototype.scheduleFaultyToTombstone = function scheduleFaultyToTombstone(member) {
+    var self = this;
+    this.schedule(member, Member.Status.faulty, this.faultyTimeout, function() {
```
mennopruijssers: style: missing function name

lib/gossip/state_transitions.js
```diff
@@ -57,6 +59,21 @@ StateTransitions.prototype.scheduleSuspectToFaulty = function scheduleSuspectToF
...
+        self.ringpop.membership.makeTombstone(member.address,
+            member.incarnationNumber);
+    });
+};
+
+StateTransitions.prototype.scheduleTombstoneToEvict = function scheduleTombstoneToEvict(member) {
+    var self = this;
+    this.schedule(member, Member.Status.tombstone, this.tombstoneTimeout, function() {
```
mennopruijssers: style: missing function name

lib/gossip/state_transitions.js
```diff
@@ -27,6 +27,8 @@ function StateTransitions(options) {
 
     var p = options.periods || {};
     this.suspectTimeout = p.suspect || StateTransitions.Defaults.suspect;
+    this.faultyTimeout = p.faulty || StateTransitions.Defaults.faulty;
```
mennopruijssers: Wouldn't it be easier to do something like this?
```
var p = _.defaults({}, options.periods, StateTransitions.Defaults);
this.suspectTimeout = p.suspect;
this.faultyTimeout = p.faulty;
this.tombstoneTimeout = p.tombstone;
``

File: lib/gossip/state_transitions.js [hunk 10]
```diff
@@ -141,7 +158,9 @@ StateTransitions.prototype.disable = function disable() {
 };
 
 StateTransitions.Defaults = {
-    suspect: 5000
+    suspect: 5 * 1000, //5s
```
mennopruijssers: style: space after slashes

lib/membership/index.js
```diff
@@ -230,6 +236,31 @@ Membership.prototype.makeSuspect = function makeSuspect(address, incarnationNumb
...
 };
 
+Membership.prototype.makeTombstone = function makeTombstone(address, incarnationNumber) {
+    this.ringpop.stat('increment', 'make-tombstone');
+    return this._updateMember(new Update(address, incarnationNumber,
+        Member.Status.tombstone, this.localMember));
+};
+
+Membership.prototype.removeMember = function removeMember(address) {
+    var hasMember = _.has(this.membersByAddress, address);
```
mennopruijssers: Should we add a safe guard against removal of self here as well?

lib/membership/index.js
```diff
@@ -230,6 +236,31 @@ Membership.prototype.makeSuspect = function makeSuspect(address, incarnationNumb
...
+    if (!hasMember) {
+        return;
+    }
+    var memberToDelete = this.membersByAddress[address];
+    delete this.membersByAddress[address];
+    this.members = _.without(this.members, memberToDelete);
+    this.computeChecksum();
+};
+
+Membership.prototype.evict = function evict(address) {
```
mennopruijssers: Most methods interacting with members are suffix with `Member` (e.g. `_updateMember`, `removeMember`, `_createMember`).
Should we change this one to `evictMember`?

lib/membership/index.js
```diff
@@ -259,6 +290,9 @@ Membership.prototype.set = function set() {
 
     for (var i = 0; i < updates.length; i++) {
         var update = updates[i];
+        if (update.status === Member.Status.tombstone) {
```
mennopruijssers: Maybe include the comment why you skip tombstones here as well?

mennopruijssers:
Some small comments but after fixing those: LGTM!

lib/gossip/state_transitions.js
thanodnl: While you are adding function names here, can you also add one to line 56?

server/admin/member.js
Motiejus: About the first comment: uh oh. Taking that back.

lib/membership/index.js
thanodnl: For some background: It has been implemented in go in two separate functions because of this discussion: https://github.com/uber/ringpop-go/pull/123#discussion_r55337319

I think it is very beneficial to have the code in go and node as close as possible to each other so it is easy to port features from 1 language to the other.

lib/membership/index.js
```diff
@@ -304,6 +338,15 @@ Membership.prototype.update = function update(changes, isLocal) {
         var member = this.findMemberByAddress(change.address);
 
         if (!member) {
+            // avoid indefinite tombstones by not creating new nodes
+            // directly in this state
+            if (change.status === Member.Status.tombstone) {
+                self.logger.info('skipping tombstone update:', {
```
thanodnl: nit: I think the colon at the end of the line is not super useful here, the fields that are after the line are not serialized in the message but added as fields on the envelope.

lib/on_membership_event.js
thanodnl: It is not `faulty` that should be here, but yes the `suspect` case should also add to the hash ring.

server/admin/member.js
```diff
@@ -99,6 +99,27 @@ function createLeaveHandler(ringpop) {
...
+        }
+
+        ringpop.membership.members.forEach(function(member) {
+            if (member.status === Member.Status.faulty) {
+                ringpop.membership.makeTombstone(member.address, member.incarnationNumber);
+            }
+        });
+
+        process.nextTick(function() {
+            callback(null, null, JSON.stringify({'status': 'ok'}));
```
thanodnl: I was thinking on adding the number of servers that were marked `tombstone`. Any thought on that, and if we should directly implement it here.

lib/on_membership_event.js
severb: Oh, that's what I meant to say; 'suspect' not 'faulty'.

server/admin/member.js
severb: Yeah, returning the number of nodes is a good idea.

lib/membership/member.js
severb: #260 Fixes this.

lib/membership/index.js
severb: I think guarding in `evict` is enough; this is just an utility.

lib/membership/index.js
severb: I think this method is closer to `makeFaulty`, `makeSuspect`, and all other `make*` methods; but because `evict` is a verb, `makeEvict` sounds awful.

server/admin/member.js
severb: https://github.com/uber/ringpop-common/pull/53

severb: I changed the integration tests to expect the count https://github.com/uber/ringpop-common/pull/53

server/admin/member.js
```diff
@@ -108,14 +108,16 @@ function createReapHandler(ringpop) {
...
+        var reaped = 0;
         ringpop.membership.members.forEach(function(member) {
             if (member.status === Member.Status.faulty) {
+                reaped++;
                 ringpop.membership.makeTombstone(member.address, member.incarnationNumber);
             }
         });
 
         process.nextTick(function() {
-            callback(null, null, JSON.stringify({'status': 'ok'}));
```
thanodnl: I think we also need `'status': 'ok'` in here, to signal that the operation succeeded. The number of reaped nodes is only extra information.

test/unit/dissemination-test.js
```diff
@@ -153,6 +153,38 @@ testRingpop('raise piggyback counter on issueAsSender', function t(deps, assert)
...
+
+    membership.makeAlive(addrAlive, incNo);
+    membership.makeSuspect(addrSuspect, incNo);
+    membership.makeFaulty(addrFaulty, incNo);
+    membership.makeTombstone(addrAlive, incNo);
+    membership.makeTombstone(addrSuspect, incNo);
+    membership.makeTombstone(addrFaulty, incNo);
+
+    // Number of expected changes
+    var expectedNumberOfChanges = membership.getMemberCount() - 1;
```
thanodnl: this variable is unused

server/admin/member.js
```diff
@@ -99,6 +99,29 @@ function createLeaveHandler(ringpop) {
...
+        var reaped = 0;
+        ringpop.membership.members.forEach(function(member) {
+            if (member.status === Member.Status.faulty) {
+                reaped++;
+                ringpop.membership.makeTombstone(member.address, member.incarnationNumber);
+            }
+        });
+
+        process.nextTick(function() {
+            callback(null, null, JSON.stringify({'reaped': reaped}));
```
thanodnl: I think we also need `'status': 'ok'` in here, to signal that the operation succeeded. The number of reaped nodes is only extra information.

thanodnl:
LGTM

severb:
@Motiejus @mennopruijssers LGTY?

mennopruijssers:
Yes. Land it!

PR #258, opened 2016-05-02
Refactor ping-sender to support sendPingWithChanges

mennopruijssers:
To support partition healing we need to manually specify the changes sent during a ping-request (instead of gossiping the changes from `dissemination`). In this case we also don't need to update the membership-list.
Both changes are required to be able to send and retrieve changes without marking all the members as faulty before merging partitions.

### DIFF
--- a/lib/gossip/ping-sender.js
+++ b/lib/gossip/ping-sender.js
@@ -61,6 +61,11 @@ PingSender.prototype.onPing = function onPing(err, res, callback) {
 PingSender.prototype.sendChanges = function sendChanges(changes, callback) {
     var self = this;
 
+    self.gossipLogger.info('ringpop ping send', {
+        local: self.ring.whoami(),
+        member: self.address,
+        changes: changes
+    });
     self.ring.client.protocolPing({
         host: self.address,
         retryLimit: self.ring.config.get('tchannelRetryLimit'),
@@ -77,7 +82,7 @@ PingSender.prototype.send = function send(callback) {
     var self = this;
 
     self.ring.dissemination.issueAsSender(function issues(changes, onIssue) {
-        self.sendChanges(changes, function onPing(err, res) {
+        self.sendChanges(changes, function changesSent(err, res) {
             onIssue(err);
             self.onPing(err, res, callback);
         });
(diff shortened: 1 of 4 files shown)


lib/gossip/ping-sender.js
```diff
@@ -57,40 +54,41 @@ PingSender.prototype.onPing = function onPing(err, res) {
...
+        self.sendChanges(changes, function onPing(err, res) {
             onIssue(err);
-            self.onPing(err, res);
+            self.onPing(err, res, callback);
         });
     });
 };
 
-module.exports = function sendPing(opts, callback) {
+PingSender.sendPing = function sendPing(opts, callback) {
```
mennopruijssers: By making this a static method on `PingSender` and exporting `PingSender` previous usage of the `sendPing`-method could just change their require from `require('ping-sender')` to `require('ping-sender').sendPing`. Partition healer will use the `sendChanges`-method in the following way:
```
var PingSender = require('ping-sender');
new PingSender(ringpop, target).sendChanges(changes, callback);
```

mennopruijssers:
`PingSender` is currently not unit-tested but covered by the integration tests in ringpop-node (and indirectly by the shared integration tests in ringpop-common as well).
However, it might be desirable to add unit-tests now or in a separate PR.

lib/gossip/ping-sender.js
```diff
@@ -57,40 +54,41 @@ PingSender.prototype.onPing = function onPing(err, res) {
...
-            checksum: self.ring.membership.checksum,
-            changes: changes,
-            source: self.ring.whoami(),
-            sourceIncarnationNumber: self.ring.membership.getIncarnationNumber()
-        }, function onPing(err, res) {
+PingSender.prototype.send = function send(callback) {
+    var self = this;
+
+    self.ring.dissemination.issueAsSender(function issues(changes, onIssue) {
+        self.sendChanges(changes, function onPing(err, res) {
```
CorgiMan: Did you remove logging? Why?

CorgiMan: The name onPing conflicts with self.onPing which is confusing to me

CorgiMan:
Looks good after you have addressed my comments.

lib/gossip/ping-sender.js
mennopruijssers: You're right, I'll add the logging again (to the `sendChanges`-method) and change the onPing as well.

PR #259, opened 2016-05-03
Refactor MemberShip~isPingable(member)

mennopruijssers:
In preparation of partition healing we need a method to check if a status is considered pingable.

### DIFF
--- a/lib/membership/member.js
+++ b/lib/membership/member.js
@@ -226,4 +226,20 @@ Member.Status = {
     suspect: 'suspect'
 };
 
+/**
+ * Check if a status is pingable. A node would be pinged if its status is either alive or suspect.
+ *
+ * @param {string} status a valid status (@see {Member.Status})
+ * @returns {boolean} true if pingable, false otherwise.
+ */
+Member.isStatusPingable = function isStatusPingable(status) {
+    switch (status) {
+        case Member.Status.alive:
+        case Member.Status.suspect:
+            return true;
+        default:
+            return false;
+    }
+};
+
 module.exports = Member;
--- a/test/unit/member_status_test.js
+++ b/test/unit/member_status_test.js
@@ -1,0 +1,45 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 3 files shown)


dansimau:
LGTM

test/unit/member_status_test.js
```diff
@@ -0,0 +1,45 @@
...
+'use strict';
+
+var _ = require('underscore');
+var test = require('tape');
+
+var Member = require('../../lib/membership/member');
+
+test('is status pingable', function(assert) {
+    var expectation = {
+        'alive': true,
```
jwolski: its always good to add a "bad" value like null/undefined/`garbage` in your tests.

test/unit/member_status_test.js
```diff
@@ -0,0 +1,45 @@
...
+        'suspect': true,
+        'faulty': false,
+        'leave': false
+    };
+
+    for(var k in Member.Status) {
```
jwolski: be careful with Javascript's `for (var ... in ...) {`. See here: http://stackoverflow.com/questions/9329446/for-each-over-an-array-in-javascript

lib/membership/member.js
```diff
@@ -226,4 +226,20 @@ Member.Status = {
     suspect: 'suspect'
 };
 
+/**
```
CorgiMan: I prefer if we keep using the same commenting style as in the rest of the project

CorgiMan:
Looks good after you address my comment.

test/unit/member_status_test.js
mennopruijssers: Yeah, that's why I'm only using it in the test-file, not in the "real code". Another option would be to manually maintain a list of all the statuses. A downside of that approach is that it's easy to forget to add a new status there. I'll add the `hasOwnProperty`-check though as a safeguard.

lib/membership/member.js
mennopruijssers: There is actually not a lot of documentation in the project. I could change it to be consistent but the nice thing about this that some IDE's actually use it and warn the user in case of misuse. In the feature we could actually generate documentation based on it (see http://usejsdoc.org/ for more info)

CorgiMan:
Looks good

PR #260, opened 2016-05-03
Simplify override logic during updates

mennopruijssers:
Based on the #259 to prevent merge-conflicts.

### DIFF
--- a/lib/membership/member.js
+++ b/lib/membership/member.js
@@ -171,68 +171,28 @@ Member.prototype._applyUpdatePenalty = function _applyUpdatePenalty() {
 };
 
 Member.prototype._isLocalOverride = function _isLocalOverride(update) {
-    var self = this;
-
-    return isLocalFaultyOverride() || isLocalSuspectOverride() || isLocalTombstoneOverride();
-
-    function isLocalFaultyOverride() {
-        return self.ringpop.whoami() === self.address &&
-            update.status === Member.Status.faulty;
+    if (this.ringpop.whoami() !== this.address) {
+        return false;
     }
 
-    function isLocalSuspectOverride() {
-        return self.ringpop.whoami() === self.address &&
-            update.status === Member.Status.suspect;
-    }
-
-    function isLocalTombstoneOverride() {
-        return self.ringpop.whoami() === self.address &&
-            update.status === Member.Status.tombstone;
-    }
+    return update.status === Member.Status.faulty ||
+        update.status === Member.Status.suspect ||
+        update.status === Member.Status.tombstone;
 };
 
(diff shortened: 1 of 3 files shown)


lib/membership/member.js
```diff
@@ -242,4 +217,26 @@ Member.isStatusPingable = function isStatusPingable(status) {
...
+            return 0;
+        case Member.Status.suspect:
+            return 1;
+        case Member.Status.faulty:
+            return 2;
+        case Member.Status.leave:
+            return 3;
+        default:
+            // unknown states will never have precedence
+            return -1;
```
CorgiMan: this actually gives the highest priority to unknown updates, unlike what the comment indicates

lib/membership/member.js
```diff
@@ -242,4 +217,26 @@ Member.isStatusPingable = function isStatusPingable(status) {
     }
 };
 
+/**
```
CorgiMan: I don't think we use this style comment in this project. I would prefer to keep it consistent.

CorgiMan:
Looks good after you address my comments, can you also add a description to this PR.

lib/membership/member.js
Motiejus: How so? I believe it's correct, and there is even a test for it below.

Motiejus:
LGTM

lib/membership/member.js
mennopruijssers: There is actually not a lot of documentation in the project. I could change it to be consistent but the nice thing about this that some IDE's actually use it and warn the user in case of misuse. In the feature we could actually generate documentation based on it (see http://usejsdoc.org/ for more info)

lib/membership/member.js
mennopruijssers: Motiejus is right: the higher the number, the higher the priority. A minus one means the opposite: lowest priority,.

CorgiMan: You are right, I was wrong

lib/membership/member.js
CorgiMan: 👍

CorgiMan:
Looks good

test/unit/member_status_test.js
```diff
@@ -24,6 +24,22 @@ var _ = require('underscore');
 var test = require('tape');
 
 var Member = require('../../lib/membership/member');
+var Ringpop = require('../../index');
+
+var ALL_STATUSES = getAllStatuses();
```
thanodnl: since we have `underscore` already as a dependency you might choose to use `_.values(Member.Status)` here instead of defining a function for 1 time use to do the same.

mennopruijssers: Didn't know about that one. Nice!

PR #261, opened 2016-05-03
Add comment about local and remote overrides

Motiejus:
This comment helps understand functions `_isLocalOverride` and `_isOtherOverride`.

### DIFF
--- a/lib/membership/member.js
+++ b/lib/membership/member.js
@@ -170,8 +170,11 @@ Member.prototype._applyUpdatePenalty = function _applyUpdatePenalty() {
     }
 };
 
-Member.prototype._isLocalOverride = function _isLocalOverride(update) {
-    var self = this;
+// Imagine I am ringpop, and I get an update. If the update is about me, then
+// it can be a localOverride. If it is a localOverride, then the update will be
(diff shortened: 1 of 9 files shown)


lib/membership/member.js
```diff
@@ -170,10 +170,13 @@ Member.prototype._applyUpdatePenalty = function _applyUpdatePenalty() {
     }
 };
 
-Member.prototype._isLocalOverride = function _isLocalOverride(update) {
-    var self = this;
+// Imagine I am ringpop, and I get an update. If the update is about me, then
+// it can be a localOverride. If it is a localOverride, then the update will be
+// applied. If it's not, then the update will be skipped.
```
mennopruijssers: This comment isn't completely correct. If `_isLocalOverride` returns true, the update will be changed to assert aliveness (and bump incarnationNumber); if not and it's also not an other overwrite, it's skipped.

mennopruijssers:
I also simplified the logic of these two methods (and make it equivalent to ringpop-go) in #260.

Motiejus:
I prefer #260 over this.

PR #263, opened 2016-05-04
Run tests in sub-directories of unit and integration as well

mennopruijssers:
Due to an incorrect glob-syntax, the tests in `tests/unit/server` were not running.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -18,4 +18,4 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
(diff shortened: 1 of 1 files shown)


Motiejus:
LGTM

dansimau:
Oops.

Just to confirm: did you verify this won't try to run any tests in `tests/ringpop-common/**`?

mennopruijssers:
@dansimau I had that directory checked out and they didn't run. Also the glob (`/{unit,integration}/**/*{_test,-test}.js`) is explicitly looking in subdirectories of `test/unit` and `test/integration`.

dansimau:
👍

PR #264, opened 2016-05-04
Partition healing

mennopruijssers:
This PR adds partition healing to ringpop-node. This initial PR is only healing manually. Periodic partition healing will be added in a subsequent PR.

### DIFF
--- a/lib/partition_healing/discover_provider_healer.js
+++ b/lib/partition_healing/discover_provider_healer.js
@@ -1,0 +1,168 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var _ = require('underscore');
+var async = require('async');
+var Member = require('../membership/member');
+var Healer = require('./healer');
+var TypedError = require('error/typed');
+
+var Errors = {
+    RingpopIsNotReadyError: require('../errors').RingpopIsNotReadyError,
+    DiscoverProviderNotAvailableError: TypedError({
+        type:'ringpop.partition-healing.discover-provider-not-available',
+        message:'discoverProvider not available to healer'
+    })
+};
+
+/**
+ *
+ *
+ * @extends Healer
+ * @param ringpop
+ * @constructor
+ */
+function DiscoverProviderHealer(ringpop) {
+    DiscoverProviderHealer.super_.call(this, ringpop);
+    this.maxNumberOfFailures = ringpop.config.get('discoverProviderHealerMaxFailures');
+}
+require('util').inherits(DiscoverProviderHealer, Healer);
+
+/**
+ * Perform a heal-operation using the discover provider (@see RingPop#discoverProvider).
+ * Briefly explained the heal-operation consists of the following steps:
+ *
+ * 1. get a target list by filtering the list of hosts from the {DiscoverProvider};
+ * 2. remove a host from the target list and try to join it;
+ * 3. merge the two membership lists when possible and gossip changes around when necessary;
+ * 4. remove the hosts that are alive according to target's membership list from the target list.
+ * 5. goto 2 until the target list is empty or the number of failures reached a configurable maximum (discoverProviderHealerMaxFailures)
+ *
+ * A full description of the algorithm is available in ringpop-common/docs.
+ *
+ * @param {Healer~healCallback} callback the callback when the heal operation is completed.
+ */
+DiscoverProviderHealer.prototype.heal = function heal(callback) {
+    var self = this;
+
+    if(!self.ringpop.isReady) {
+        callback(Errors.RingpopIsNotReadyError());
+        return;
+    }
+    if (!self.ringpop.discoverProvider) {
+        var error = Errors.DiscoverProviderNotAvailableError();
+        self.logger.warn(error.message, {
+            local: self.ringpop.whoami()
+        });
+        callback(error);
+        return;
+    }
+
+    self.ringpop.discoverProvider(onHostsDiscovered);
+
+    function onHostsDiscovered(err, hosts) {
+        if (err) {
+            self.logger.warn('ringpop unable to retrieve host list from discover provider during heal', {
+                local: self.ringpop.whoami(),
+                err: err
+            });
+            return callback(err);
+        }
+
+        var potentialTargets = self._getTargets(hosts);
+        var targets = [];
+        var numberOfFailures = 0;
+
+        async.whilst(
+            function shouldContinueHealing() {
+                return potentialTargets.length > 0 && numberOfFailures < self.maxNumberOfFailures;
+            },
+            function healTargetIterator(next) {
+                var target = potentialTargets.pop();
+
+                self.attemptHeal(target, function onHealAttempt(err, reachableNodes) {
+                    if (err) {
+                        numberOfFailures++;
+                        self.logger.warn('ringpop heal attempt failed', {
+                            local: self.ringpop.whoami(),
+                            target: target,
+                            numberOfFailures: numberOfFailures,
+                            maxNumberOfFailures: self.maxNumberOfFailures,
+                            err: err
+                        });
+
+                        // continue
+                        next();
+                        return;
+                    }
+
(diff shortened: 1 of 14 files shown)


mennopruijssers:
Currently based on a "special" branch: master merged with #260 and #259

mennopruijssers:
Manual test:
```
# create 25 node cluster
./scripts/tick-cluster.js -n 25 --interpreter node main.js

# validate no partitions:
ringpop-admin partitions $IP:3010
 08:59:18.496   Checksum    # Nodes   # Alive   # Suspect   # Faulty   Sample Host
                140365291   25        25        0           0          $IP:3010

# setup firewall using make_partition-script in ringpop-common:
sudo lsof -Pnni | ./make_partition 3000-3010 3011-3024  | sudo pfctl -emf -

# reset firewall after a while:
sudo pfctl -f /etc/pf.conf

# validate two partitions:
ringpop-admin partitions $IP:3010
 09:01:09.012   Checksum    # Nodes   # Alive   # Suspect   # Faulty   Sample Host
                228740776   14        14        0           11         $IP:3011
                883647694   11        11        0           14         $IP:3010

# run heal:
ringpop-admin heal $IP:3000
09:02:11.273 Executed heal to 1 targets
09:02:11.273  - $IP:3011
09:02:12.285 Executed heal to 1 targets
09:02:12.285  - $IP:3022
09:02:13.286 No known partitions left

# validate no partitions:
ringpop-admin partitions $IP:3010
 09:02:14.495   Checksum     # Nodes   # Alive   # Suspect   # Faulty   Sample Host
                1366565648   25        25        0           0          $IP:3010

```

Motiejus:
Please make sure to run integration tests in travis-ci too. Appending `--enable partition-healing` to the `it-tests.js` command should do the trick here: https://github.com/uber/ringpop-node/blob/dc57cbef6a8b377daa9d24f43b25fdc20be6a5c4/test/run-shared-integration-tests#L81

lib/partition_healing/healer.js
```diff
@@ -0,0 +1,193 @@
...
+            // if the local member would become un-pingable after applying the remote change
+            if (nodeWouldBecomeUnPingable(localMember, remoteMember)) {
+                // mark it locally as suspect
+                localChanges.push(_.extend({}, remoteMember, {status: Member.Status.suspect}));
+            }
+
+            // if the remote member would become un-pingable after applying the local change
+            if (nodeWouldBecomeUnPingable(remoteMember, localMember)) {
+                // mark it remotely as suspect
+                remoteChanges.push(_.extend({}, localMember, {status: Member.Status.suspect}));
```
thanodnl: since the local changes are created from the `fullsync` call they contain `source` information about this node if I remember correctly. This in combination with bidirectional fullsync's caused an issue in the ringpop-node implementation.

I think we solved it by scrubbing (or never putting it on there) the source information from the change we sent remote.

If we don't do so the remote might initiate a fullsync back to this node when it receives the 'reincarnation' ping during the partition healing causing the 'bridge' between the two partitions to be created too soon.

lib/partition_healing/discover_provider_healer.js
```diff
@@ -63,6 +63,7 @@ require('util').inherits(DiscoverProviderHealer, Healer);
  */
 DiscoverProviderHealer.prototype.heal = function heal(callback) {
     var self = this;
+    self.ringpop.stat('increment', 'heal.triggered.discover_provider');
```
thanodnl: the stat we emit here differs from the stat in `ringpop-go`: https://github.com/uber/ringpop-go/blob/dev/ringpop.go#L497 is that on purpose?

test/integration/partition_healing_test.js
```diff
@@ -0,0 +1,142 @@
...
+        for (var i = 0; i < partitionA.length; i++) {
+            var node = partitionA[i];
+            for (var j = 0; j < node.membership.members.length; j++) {
+                var member = node.membership.members[j];
+                if (_.pluck(partitionA, 'hostPort').indexOf(member.address) > -1) {
+                    assert.equal(member.status, 'alive')
+                } else if (_.pluck(partitionB, 'hostPort').indexOf(member.address) > -1) {
+                    assert.equal(member.status, 'faulty');
+                } else {
+                    assert.fail('!!!!');
```
thanodnl: can we put a better description for the failure in the string? What is actually wrong when we reach this branch? It would make the output of the test more understandable as well as reading this test.

test/run-shared-integration-tests
```diff
@@ -78,7 +78,7 @@ run-test-for-cluster-size() {
     # Run the tests and buffer the output to a log file. We'll display it later
     # if the test fails. This avoids interleaving of output to the terminal
     # when tests are running in parallel.
-    node "${ringpop_common_dir}/test/it-tests.js" \
+    node "${ringpop_common_dir}/test/it-tests.js" --enable-feature partition-healing \
```
thanodnl: nit: for easier to understandable diffs now and in the future I think it is easiest if we put `--enable-feature` flags here on separate lines. It will cause only 1 addition instead of a remove and an addition when adding tests. Same when we can remove features because they became mandatory.

lib/partition_healing/discover_provider_healer.js
mennopruijssers: Good point. I wanted to discuss this with @CorgiMan.
Since most partition healing code is already prepared for multiple strategies, I thought it made more sense to increment a more specific counter. @CorgiMan do you agree?

CorgiMan: I think that's fine, but let's keep the stats the same by either changing go to this style, or changing node to go's style/

mennopruijssers: See uber/ringpop-go#142

lib/partition_healing/healer.js
```diff
@@ -0,0 +1,208 @@
...
+            // already un-pingable
+            return false;
+        }
+
+        if (Member.isStatusPingable(newState.status)) {
+            // new state is pingable
+            return false;
+        }
+
+        if (currentState.incarnationNumber > newState.incarnationNumber) {
```
CorgiMan: Isn't this logic done by, isNonLocalOverwrite?

CorgiMan: Nevermind

lib/partition_healing/discover_provider_healer.js
```diff
@@ -0,0 +1,169 @@
...
+                    }
+
+                    targets.push(target);
+                    // Remove all reachable nodes from the list of potential targets.
+                    potentialTargets = _.difference(potentialTargets, reachableNodes);
+                    next();
+                });
+            },
+            function healingDone(err) {
+                if (numberOfFailures >= self.maxNumberOfFailures) {
```
Motiejus: Do you think it would make sense to add a stat for that?

lib/partition_healing/healer.js
```diff
@@ -0,0 +1,208 @@
...
+ */
+function Healer(ringpop) {
+    this.ringpop = ringpop;
+    this.logger = this.ringpop.loggerFactory.getLogger('healer');
+}
+
+
+/**
+ * Attempt a heal between the current and target node. Calling this function does not
+ * result in a heal ff there are nodes that need to be reincarnated to take precedence
```
Motiejus: s/ff/iff/ ?

lib/partition_healing/healer.js
```diff
@@ -0,0 +1,208 @@
...
+            source: self.ringpop.whoami(),
+            incarnationNumber: self.ringpop.membership.localMember.incarnationNumber
+        }, next);
+    }
+
+    function generateChanges(joinResponse, next) {
+        remoteMembership = joinResponse.membership;
+
+        // Index local membership by address for faster lookups.
+        var localMembershipMap = _.indexBy(self.ringpop.dissemination.fullSync(), 'address');
```
Motiejus: In the go version, we renamed `fullSync` (equivalent to what we have here) to `MembershipAsChanges`, because it's not a full-sync anymore. Should we rename this too?

lib/partition_healing/healer.js
```diff
@@ -0,0 +1,208 @@
...
+
+        var localChanges = [];
```
Motiejus: In the RFCs, partition healing documentation, and the public documentation, `local*` and `remote*` are called `A*` and `B*` respectively. In go, we have `changesForA` and `changesForB`.

Current way it feels a bit confusing, because just a few lines above we have `localMembershipMap`, which has a different purpose.

What do you think about renaming these two to match the Go implementation?

lib/partition_healing/healer.js
```diff
@@ -0,0 +1,208 @@
...
+                // mark it remotely as suspect
+                remoteChanges.push(createSuspectChange(localMember));
+            }
+        }
+
+        next(null, localChanges, remoteChanges);
+    }
+
+    function createSuspectChange(member) {
+        // ignore source and sourceIncarnationNumber fields to prevent bi-directional full sync
```
Motiejus: Did you mean "do not send" instead of "ignore"?

mennopruijssers: Yeah that might be a better comment.

lib/partition_healing/healer.js
mennopruijssers: will do.

mennopruijssers: I though this was actually easier to read because localChanges are applied to the local change and remoteChanges to the target (so the remote node) but it might make sense to keep them in sync.

lib/partition_healing/healer.js
mennopruijssers: We could do it but we should do it in another pull request to keep this one as small as possible.

lib/partition_healing/discover_provider_healer.js
mennopruijssers: There isn't a stat in ringpop-go. We might add it here but we have to add it to ringpop-go as well. @CorgiMan what do you think?

lib/partition_healing/healer.js
mennopruijssers: @Motiejus #266

lib/partition_healing/healer.js
mennopruijssers: Decided to make them in sync in the end. The only difference is that I named MA and MB membershipA and membershipB.

lib/partition_healing/discover_provider_healer.js
CorgiMan: Yeah sounds like a good idea.

mennopruijssers: Cool, I'll create a separate task (and PR) for it (also for ringpop-go).

Motiejus:
LGTM

thanodnl:
LGTM, but make sure tests pass :)

PR #265, opened 2016-05-09
Periodical partition healing

mennopruijssers:
Try to heal partitions periodically. By using a probability variable, on average `x` amount of nodes will try to perform the heal every `y` seconds (where `x` = `discoverProviderHealerBaseProbability`, and `y` = `discoverProviderHealerPeriod`). 
The defaults are 3 nodes every 30 seconds.

### DIFF
--- a/lib/partition_healing/discover_provider_healer.js
+++ b/lib/partition_healing/discover_provider_healer.js
@@ -22,6 +22,7 @@
 
 var _ = require('underscore');
 var async = require('async');
+var globalTimers = require('timers');
 var Member = require('../membership/member');
 var Healer = require('./healer');
 var TypedError = require('error/typed');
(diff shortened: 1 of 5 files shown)


lib/partition_healing/discover_provider_healer.js
```diff
@@ -43,10 +44,63 @@ var Errors = {
...
+        this.timers.clearTimeout(this.healTimer);
+        this.healTimer = null;
+    }
+};
+
+DiscoverProviderHealer.prototype._run = function _run() {
+    var self = this;
+
+    if (self.isStopped) {
+        return;
```
severb: I think a log message is in place here.

lib/partition_healing/discover_provider_healer.js
```diff
@@ -43,10 +44,63 @@ var Errors = {
...
+    }
+
+    if (Math.random() < probability()) {
+        self.heal(function onHeal() {
+            scheduleNext();
```
severb: Any reason, other than readability, for why this can't be `self.heal(scheduleNext)`?

severb:
Nice and readable code, LGTM. I also like the small PR approach, I think it pays off.

lib/partition_healing/discover_provider_healer.js
mennopruijssers: Pure readability: otherwise it's easier to miss the `scheduleNext`-call when reading the code.

mennopruijssers:
@severb Thanks!

thanodnl:
LGTM

PR #266, opened 2016-05-10
Rename Dissemination.fullSync to membershipAsChanges

mennopruijssers:
In the ringpop-go we renamed fullSync to MembershipAsChanges. This pull-request is doing the same thing in ringpop-node because that's better describing what the function does: returning the full membership as changes.


Motiejus:
LGTM

PR #268, opened 2016-05-13
Don't use async library in production-code

mennopruijssers:
Partition healing was by accident using `async`: a dev-dependency. Let's not add an extra dependency and do it without `async`!


mennopruijssers:
Closed since the offending commits are reverted in #270.

PR #269, opened 2016-05-17
Prune dependencies when running integration tests

mennopruijssers:
Recently we used a dev-dependency in production code by accident. By pruning the dependencies to only use production dependencies during integration tests we make sure this doesn't happen again.
However, while implementing it I discovered a problem with main.js: it requires `commander` and `tchannel` which are both listed as dev-dependencies.


mennopruijssers:
(travis will fail until #268 is merged into master)

package.json
```diff
@@ -60,13 +61,15 @@
...
     "tape-cluster": "2.1.0",
-    "tchannel": "^3.6.13",
     "tcurl": "^4.11.1",
     "timer-shim": "^0.3.0",
     "tryit": "^1.0.1",
     "uber-licence": "^2.0.1",
     "winston": "^1.0.1"
   },
+  "peerDependencies": {
+    "tchannel": "^3.6.13"
```
mennopruijssers: Since a `tchannel`-instance is passed into `ringpop` during construction, it should be listed as a peer dependency so we can explicitly state our supported version when a library is using ringpop as a dependency.

package.json
```diff
@@ -31,10 +31,12 @@
   },
   "dependencies": {
     "body": "^5.0.0",
+    "commander": "^2.6.0",
```
thanodnl: I like the idea of pruning the dependencies to production, I don't really like to have commander now as a dependency. This is only used when ringpop is ran during testing right? Is there a way to have separate testing dependencies separate from `devDependencies`

mennopruijssers: No that's not possible. The only other way I could think of is to create a separate npm-module (e.g. testpop) in a subdirectory that requires ringpop using a relative path. 

Also, main.js is currently installed as the `ringpop`-binary (see https://github.com/uber/ringpop-node/blob/master/package.json#L17) which means that ringpop currently not only depends on commander during testing. So we should either add commander as a dependency or stop providing the ringpop-binary.

package.json
```diff
@@ -13,9 +13,6 @@
   ],
   "version": "10.14.0",
   "repository": "git://github.com/uber/ringpop.git",
-  "bin": {
```
mennopruijssers: I could move this change to a separate PR if preferred.

thanodnl:
LGTM

dansimau:
Not sure I totally agree with the approach here.

* Pruning devDependencies during integration tests seems to make sense to me. 👍 It's an added check we can do to prevent relying on a dev dependency in production and would prevent the issue we had the other day.
* If `main.js` ("testpop") is part of ringpop-node, then I think we should include it's dependencies in the `dependencies` section of `package.json` (and not `devDependencies`, where it curently is); because a ringpop-node binary _genuinely_ depends on it.
* However because we're a library, we should be smart about this: to avoid our customers from having to install tchannel and commander then we should move our "testpop" binary to a subdir with it's own `package.json`. This is what ringpop-go does.
* If we do that, the integration test runner just needs a build step for testpop. This matches ringpop-go.

I **do not agree** with adding a new whitelist to the integration test runner script itself. This seems like an obscure place to list yet another set of dependencies and I can see someone getting tripped up by this in the future.

package.json
Motiejus: IIRC `main.js` is used only during development (run it manually), so I think this doesn't need to be a binary.

+1 for making commander a devDependency.

mennopruijssers:
Closing since we need a better way of doing this.

PR #271, opened 2016-05-17
Emit stats from ringpop-test program

mennopruijssers:
In preparation for automated performance and failure testing we want to be able to emit stats to a file or a network address from testpop.
This commit adds a new flag to `main.js` to control if and where stats should be emitted; if the flag is omitted, no stats are emitted.


main.js
```diff
@@ -68,6 +79,69 @@ function main(args) {
...
+    } else {
+        var file = path.resolve(val);
+        opts = {
+            _ephemeralSocket: new FileStatsLogger(file)
+        };
+    }
+
+    var createStatsdClient;
+
+    try {
```
mennopruijssers: By using a `try/catch` here we're able to use `uber-statsd-client` without adding it as a dependency.

main.js
```diff
@@ -68,6 +79,71 @@ function main(args) {
     }
 }
 
+function createStatsClient(val) {
+    var opts = null;
+    var matchesHostPort = val.match(/^(.+):(\d+)$/);
```
Motiejus: I believe this approach to parse this string is quite brittle; see my comment on https://github.com/uber/ringpop-go/pull/144/files#r64207110.

Motiejus:
Same question as in [Go version][1]. How are we going to test this?

[1]: https://github.com/uber/ringpop-go/pull/144#issuecomment-220955239

mennopruijssers:
@Motiejus currently main.js is currently not tested in an automated way: it's just a simple "example"-app. If we want to test this, I would vote for moving it to it's own module or repo, let it depend on uber/ringpop-node#master and add tests there for behaviour like this (and do the same for go's counterpart). I don't think that it should be part of this PR but could be a "pre-requirement" if we decide to move forward with this approach.

Motiejus:
I think we can make at least a unit test for the matching part; shouldn't be too hard?

I still think overloading a single configuration key for two different values doesn't seem right; do you think otherwise (we can have a talk, maybe I`m misunderstanding something)

Motiejus:
LGTM!

main.js
```diff
@@ -47,6 +50,15 @@ function main(args) {
             'The lifetime of a tombstone member in ms. After that the member is removed from the membership.',
             parseInt10, 5000)
 
+        .option('--stats-file <stats-file>',
+            'Enable stats emitting to a file. Stats-file can be a relative or absolute path. '+
+            'Note: this flag is mututally excslusive with --stats-udp and you need to manually install "uber-statsd-client" to be able to emit stats')
```
btromanova: Do you need uber-statsd-client to emit to file?

main.js
```diff
@@ -47,6 +50,15 @@ function main(args) {
...
+
+        .option('--stats-udp <stats-udp>',
+            'Enable stats emitting over udp. Destination is in the host-port format (e.g. localhost:8125 or 127.0.0.1:8125) ' +
+            'Note: this flag is mututally excslusive with --stats-udp and you need to manually install "uber-statsd-client" to be able to emit stats',
```
btromanova: should be "mutually exclusive with --stats-file", I guess

main.js
```diff
@@ -87,6 +102,80 @@ function main(args) {
...
+    if (program.statsUdp) {
+        var matchesHostPort = program.statsUdp.match(/^(.+):(\d+)$/);
+        opts = {
+            host: matchesHostPort[1],
+            port: parseInt(matchesHostPort[2])
+        };
+    } else if (program.statsFile) {
+        var file = path.resolve(program.statsFile);
+        opts = {
+            _ephemeralSocket: new FileStatsLogger(file)
```
btromanova: Can you please add comments on how it works?

btromanova:
After small fix in option description and comment about using _ephemeralSocket --- lgtm.

PR #273, opened 2016-05-20
Change emitted stat name to heal.triggered

btromanova:
Previous value was heal.triggered.discover_provider. Changing it to be in sync with ringpop-go.


mennopruijssers:
LGTM

CorgiMan:
Looks good!

PR #274, opened 2016-05-24
Fix infinite loop in tick-cluster kill and suspend.

btromanova:
tick-cluster would currently go into infinite loop if I'll try to kill or suspend more processes, then are running. The new approach is to filter only proc's that are running (not suspended or killed), shuffle them and kill/suspend first min(requestedTo, running.lenght).

### DIFF
--- a/scripts/tick-cluster.js
+++ b/scripts/tick-cluster.js
@@ -21,6 +21,7 @@
 
 'use strict';
 
+var _ = require('underscore');
 var childProc = require('child_process');
 var color = require('cli-color');
 var generateHosts = require('./generate-hosts');
@@ -29,7 +30,7 @@ var TChannel = require('tchannel');
 var fs = require('fs');
 
 var programInterpreter, programPath, startingPort, bindInterface, procsToStart = 5;
-var hosts, procs, ringPool, localIP, toSuspend, toKill, tchannel; // defined later
+var hosts, procs, ringPool, localIP, tchannel; // defined later
 
 /* jshint maxparams: 6 */
 
@@ -52,16 +53,6 @@ function lpad(num, len) {
 function formatDate() {
     var now = new Date();
     return lpad(now.getHours(), 2) + ':' + lpad(now.getMinutes(), 2) + ':' + lpad(now.getSeconds(), 2) + '.' + lpad(now.getMilliseconds(), 3);
-}
-
-function shuffle(a) {
-    for (var i = a.length - 1; i > 0; i--) {
-        var j = Math.floor(Math.random() * (i + 1));
-        var tmp = a[i];
-        a[i] = a[j];
-        a[j] = tmp;
-    }
-    return a;
 }
 
 function logMsg(who, msg) {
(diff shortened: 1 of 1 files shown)


scripts/tick-cluster.js
```diff
@@ -54,6 +54,16 @@ function formatDate() {
     return lpad(now.getHours(), 2) + ':' + lpad(now.getMinutes(), 2) + ':' + lpad(now.getSeconds(), 2) + '.' + lpad(now.getMilliseconds(), 3);
 }
 
+function shuffle(a) {
```
mennopruijssers: One of the dependencies (`underscore`) already implements a `shuffle`-function. It would make more sense to use that one instead of implementing our own.

scripts/tick-cluster.js
```diff
@@ -429,34 +439,32 @@ function reviveProcs() {
...
-        var proc = procs[rand];
-        if (proc.killed === null && proc.suspended === null) {
-            logMsg(proc.port, color.green('pid ' + proc.pid) + color.red(' randomly selected for sleep'));
-            process.kill(proc.proc.pid, 'SIGSTOP');
-            proc.suspended = Date.now();
-            suspended.push(proc);
-        }
+    var running = procs.filter(function (proc) { return !proc.killed && !proc.suspended; });
+    toSuspend = Math.min(+count, running.length);
+    shuffle(running);
```
mennopruijssers: Instead of using underscore's shuffle method, you could also use `sample`: http://underscorejs.org/#sample.
```
processesToSuspend = _.sample(running, +count)
processesToSuspend.each(function suspend(proc) {
    logMsg(proc.port, color.green('pid ' + proc.pid) + color.red(' randomly selected for sleep'));
    process.kill(proc.proc.pid, 'SIGSTOP');
    proc.suspended = Date.now();
}
```

scripts/tick-cluster.js
```diff
@@ -429,34 +439,32 @@ function reviveProcs() {
...
-        var proc = procs[rand];
-        if (proc.killed === null && proc.suspended === null) {
-            logMsg(proc.port, color.green('pid ' + proc.pid) + color.red(' randomly selected for death'));
-            process.kill(proc.proc.pid, 'SIGKILL');
-            proc.killed = Date.now();
-            killed.push(proc);
-        }
+    var running = procs.filter(function (proc) { return !proc.killed && !proc.suspended; });
+    toKill = Math.min(+count, running.length);
+    shuffle(running);
```
mennopruijssers: Same here

scripts/tick-cluster.js
mennopruijssers: You might even leverage chaining to make it a "one-liner":
```
var processesToSuspend = _.chain(procs)
    .filter(function (proc) { return !proc.killed && !proc.suspended; })
    .sample(+count)
    .value();

btromanova: Thanks for the suggestion, could you please have another look?

mennopruijssers:
LGTM!

mennopruijssers:
might be awesome if you could merge your changes into ringpop-common as well (https://github.com/uber/ringpop-common/blob/master/tools/tick-cluster.js)

btromanova:
@mennopruijssers thanks for the last comment. I'll try to remove tick-cluster from ringpop-node to be able to keep one version in ringpop-common.

btromanova:
tick-cluster was moved to ringpop-common, the fix is implemented there.

PR #277, opened 2016-05-26
Read hosts-file async

mennopruijssers:
Since discover-providers are using a callback to return the host list, the hosts-file could be able to read the file asynchronously.

### DIFF
--- a/discover-providers.js
+++ b/discover-providers.js
@@ -53,11 +53,6 @@ function createJsonFileDiscoverProvider(hostsFile) {
     var fs = require('fs');
 
     return function jsonFileProvider(callback) {
-        if (!fs.existsSync(hostsFile)) {
-            callback(new Error('bootstrap hosts file does not exist', {file: hostsFile}));
-            return;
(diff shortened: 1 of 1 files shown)


discover-providers.js
```diff
@@ -58,15 +58,21 @@ function createJsonFileDiscoverProvider(hostsFile) {
...
+        fs.readFile(hostsFile, function onFileRead(err, data) {
+            if (err) {
+                callback(err);
+                return;
+            }
 
-        return callback(null, hosts);
+            try {
+                var hosts =JSON.parse(data.toString());
+                callback(null, hosts);
```
thanodnl: if the callback throws an error it is caught by the try catch resulting in an extra call of the `callback`.

Can you move the invocation outside the `try`?

mennopruijssers: good point, I'll change it!

thanodnl:
LGTM

CorgiMan:
LGTM

PR #278, opened 2016-05-27
Remove generate-hosts and tick-cluster scripts.

btromanova:
Scripts were moved to ringpop-common, because they are used for integration tests in multiple projects: ringpop-go, ringpop-admin and ringpop-node.


dansimau:
LGTM!

README.md
```diff
@@ -25,12 +25,13 @@ ulimit -n 4096
 ```
 
 # Tick Cluster
-An example application `scripts/tick-cluster.js` is included to this
+An example application `tools/tick-cluster.js` is included to ringpop-common
```
mennopruijssers: s/included to/included in

jwolski:
lgtm

PR #280, opened 2016-05-30
add suspect, faulty and tombstone period flags

CorgiMan:
Add suspect, faulty and tombstone period flags so that integration tests can set a small faulty->tombstone period of 5 seconds while having a big faulty->tombstone period when running from tickcluster.

### DIFF
--- a/main.js
+++ b/main.js
@@ -41,9 +41,9 @@ function main(args) {
         process.exit(1);
     }
 
-    program.suspect = parseInt(program.suspect) || 5;
-    program.faulty = parseInt(program.faulty) || 24*60*60;
-    program.tombstone = parseInt(program.tombstone) || 5;
+    program.suspect = parseInt(program.suspect, 10) || 5;
+    program.faulty = parseInt(program.faulty, 10) || 24*60*60;
+    program.tombstone = parseInt(program.tombstone, 10) || 5;
 
     var tchannel = new TChannel({
     });
--- a/main.js
+++ b/main.js
@@ -29,6 +29,9 @@ function main(args) {
         .usage('[options]')
         .option('-l, --listen <listen>', 'Host and port on which server listens (also node\'s identity in cluster)')
         .option('-h, --hosts <hosts>', 'Seed file of list of hosts to join')
+        .option('--suspect [suspect]', 'Suspect period in seconds')
+        .option('--faulty [faulty]', 'Faulty period in seconds')
+        .option('--tombstone [tombstone]', 'Tombstone period in seconds')
         .parse(args);
 
     var listen = program.listen;
@@ -37,6 +40,10 @@ function main(args) {
         program.outputHelp();
         process.exit(1);
     }
+
+    program.suspect = parseInt(program.suspect) || 5;
+    program.faulty = parseInt(program.faulty) || 24*60*60;
+    program.tombstone = parseInt(program.tombstone) || 5;
 
     var tchannel = new TChannel({
     });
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -29,6 +29,9 @@ function main(args) {
         .usage('[options]')
         .option('-l, --listen <listen>', 'Host and port on which server listens (also node\'s identity in cluster)')
         .option('-h, --hosts <hosts>', 'Seed file of list of hosts to join')
+        .option('--suspect [suspect]', 'Suspect period in seconds')
```
Motiejus: I think these should be `suspect-period`, `faulty-period` and `tombstone-period` respectively, since, at least I am used to noun being a boolean (binary) meaning.

main.js
```diff
@@ -38,6 +41,10 @@ function main(args) {
         process.exit(1);
     }
 
+    program.suspect = parseInt(program.suspect, 10) || 5;
+    program.faulty = parseInt(program.faulty, 10) || 24*60*60;
+    program.tombstone = parseInt(program.tombstone, 10) || 5;
```
Motiejus: Did you reverse faulty and tombstone?

main.js
```diff
@@ -38,6 +41,10 @@ function main(args) {
         process.exit(1);
     }
 
+    program.suspect = parseInt(program.suspect, 10) || 5;
```
mennopruijssers: You can also use:
```
.option('--suspect <suspect>', 'Suspect period in seconds', parseInt, 5)
``
to parse to int and specify a default in the definition of the option.

File: main.js (hunk 1 above)
Comment (mennopruijssers) [inline main.js]: you should use `--suspect  <suspect>` to make a value required if `--suspect` is specified.

File: main.js [hunk 4]
```diff
@@ -51,8 +58,9 @@ function main(args) {
         }),
         isCrossPlatform: true,
         stateTimeouts: {
-            faulty: 5 * 1000, // 5s
-            tombstone: 5 * 1000 // 5s
+            suspect: program.suspect * 1000,
```
mennopruijssers: You could even instead of parseInt, specify a custom function to the `.option`-call that does a parseInt * 1000 for you:
```
function secondsStringToMilliseconds(s) {
    return parseInt(s, 10) * 1000;
}
...
.option('--suspect <suspect>', 'Suspect period in seconds', secondsStringToMilliseconds, 5)
```

main.js
CorgiMan: No, the time from faulty to tombstone is 24*60*60

main.js
```diff
@@ -29,9 +29,9 @@ function main(args) {
         .usage('[options]')
         .option('-l, --listen <listen>', 'Host and port on which server listens (also node\'s identity in cluster)')
         .option('-h, --hosts <hosts>', 'Seed file of list of hosts to join')
-        .option('--suspect [suspect]', 'Suspect period in seconds')
-        .option('--faulty [faulty]', 'Faulty period in seconds')
-        .option('--tombstone [tombstone]', 'Tombstone period in seconds')
+        .option('--suspect-period <suspectPeriod>', 'Suspect period in ms', parseInt10, 5000)
```
mennopruijssers: parseInt10 is not a global function in node (but parseInt is exactly what you want).

mennopruijssers:
Same comment on the descriptions as in the go version (uber/ringpop-go#148) but besides that: LGTM.

mennopruijssers:
Owh, there is a linter error though: 
```
main.js: line 76, col 29, Missing semicolon. (W033)
```

CorgiMan:
Anything else?

Motiejus:
LGTM

PR #281, opened 2016-06-07
Emit server-added and server-removed stat on ringChanged event

btromanova:
Stat is emitted on ringChange in go-version, need to keep node version in-sync.


btromanova:
Similar logic in go: https://github.com/uber/ringpop-go/blob/dev/ringpop.go#L509

Versions will still be not in sync in emitting events after this change, because node version emits ringServerAdded events on hashring.addServer (same for remove) and go version emits RingChangedEvent. But I think it's a separate issue and can be fixed later.

mennopruijssers:
LGTM, can you add tests for this though?

CorgiMan:
Looks good!

PR #282, opened 2016-06-07
When creating the change for reincarnation use your self as the source.

thanodnl:
When a gossip was received that would overwrite the state of the local node we overwrite this change with a reincarnated version of our self. However we used the source of the original change as the source for this change making it harder for the reincarnated change to reach the node that declared it a suspect in the first place.

This change will use the source and new reincarnation number of the local node for this new change. This prevents the reincarnation change to be filtered out in pings to the original source causing convergence to speedup and potentially prevent fullsyncs from happening.

### DIFF
--- a/lib/membership/member.js
+++ b/lib/membership/member.js
@@ -94,7 +94,7 @@ Member.prototype.evaluateUpdate = function evaluateUpdate(update) {
         update = {
             source: this.ringpop.whoami(),
             sourceIncarnationNumber: newIncNumber,
-            address: this.id,
+            address: this.address,
             status: Member.Status.alive,
             incarnationNumber: newIncNumber
         };
--- a/lib/membership/member.js
+++ b/lib/membership/member.js
@@ -20,7 +20,6 @@
 
 'use strict';
 
-var _ = require('underscore');
 var EventEmitter = require('events').EventEmitter;
 var events = require('./events.js');
 var numOrDefault = require('../util.js').numOrDefault;
@@ -91,10 +90,14 @@ Member.prototype.evaluateUpdate = function evaluateUpdate(update) {
     if (this._isLocalOverride(update)) {
         // Override intended update. Assert aliveness!
         this.ringpop.stat('increment', 'refuted-update');
-        update = _.defaults({
+        var newIncNumber = this.Date.now();
+        update = {
+            source: this.ringpop.whoami(),
+            sourceIncarnationNumber: newIncNumber,
+            address: this.id,
             status: Member.Status.alive,
-            incarnationNumber: this.Date.now()
-        }, update);
+            incarnationNumber: newIncNumber
+        };
     } else if (!this._isOtherOverride(update)) {
         return;
     }
--- a/test/unit/membership_test.js
+++ b/test/unit/membership_test.js
@@ -75,12 +75,15 @@ testRingpop('change with higher incarnation number results in leave override', f
 testRingpop('change that overrides the local status should be overwritten to a change that reincarnates the node', function t(deps, assert) {
     var ringpop = deps.ringpop;
     var membership = deps.membership;
+    var source = "192.0.2.1:1234";
+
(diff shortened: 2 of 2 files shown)


lib/membership/member.js
```diff
@@ -91,10 +90,14 @@ Member.prototype.evaluateUpdate = function evaluateUpdate(update) {
...
+        var newIncNumber = this.Date.now();
+        update = {
+            source: this.ringpop.whoami(),
+            sourceIncarnationNumber: newIncNumber,
+            address: this.id,
             status: Member.Status.alive,
-            incarnationNumber: this.Date.now()
-        }, update);
+            incarnationNumber: newIncNumber
+        };
```
Motiejus: Since you are gossiping this message downstream, I think it is still reasonable to still base it off of the original one changing the fields you need to change, rather than reconstruct. Will be good for forwards compatibility if we decide to add a new field in the future (similarly like google protocol buffers preserve fields that it doesn't understand when forwarding a message).

test/unit/membership_test.js
```diff
@@ -72,6 +72,33 @@ testRingpop('change with higher incarnation number results in leave override', f
...
 
+testRingpop('change that overrides the local status should be overwritten to a change that reincarnates the node', function t(deps, assert) {
+    var ringpop = deps.ringpop;
+    var membership = deps.membership;
+
+    var member = membership.findMemberByAddress(ringpop.whoami());
+    assert.equals(member.status, Member.Status.alive, 'member starts alive');
+
+    var applied = membership.update([{
+        source: "192.0.2.1:1234",
```
Motiejus: Can we assert somewhere in the test that `"192.0.2.1:1234" != ringpop.whoami()` to make sure line 98 is testing what we expect?

CorgiMan:
Looks good after you handled Motiejus' comments.

lib/membership/member.js
mennopruijssers: I'm not sure I agree: a update that's causing a local override should result in a _new_ update gossiped around marking it as alive again. A node should never create a new update with fields it don't understand to be sure to not break forward compatibility.
example: let's say we add an uuid field on updates in the future. Just changing the stuff we want would result in two contradicting updates with the same uuid where nodes could easily preserve backwards compatibility by checking if an uuid is present or not.

lib/membership/member.js
```diff
@@ -91,10 +90,14 @@ Member.prototype.evaluateUpdate = function evaluateUpdate(update) {
     if (this._isLocalOverride(update)) {
         // Override intended update. Assert aliveness!
         this.ringpop.stat('increment', 'refuted-update');
-        update = _.defaults({
+        var newIncNumber = this.Date.now();
+        update = {
+            source: this.ringpop.whoami(),
+            sourceIncarnationNumber: newIncNumber,
+            address: this.id,
```
mennopruijssers: `address: this.address`?

lib/membership/member.js
```diff
@@ -91,10 +90,14 @@ Member.prototype.evaluateUpdate = function evaluateUpdate(update) {
     if (this._isLocalOverride(update)) {
         // Override intended update. Assert aliveness!
         this.ringpop.stat('increment', 'refuted-update');
-        update = _.defaults({
+        var newIncNumber = this.Date.now();
+        update = {
+            source: this.ringpop.whoami(),
```
mennopruijssers: In `update.js` the `address` field of a member is used as the source.

lib/membership/member.js
Motiejus: I guess we can look at it both ways (forwards compatibility versus risk of not changing fields that should change), and I don't have a strong preference here. So feel free to do either way.

lib/membership/member.js
thanodnl: I see, however, in `member.js` we have no access to the `localMember` which is accessible in `update.js`. In the `_isLocalOverride` function that tests if the update is about this node the `whoami` function is used as well.

mennopruijssers:
Nice! LGTM

Motiejus:
LGTM

PR #284, opened 2016-06-09
Fix double reincarnation

thanodnl:
This change will add a check for the incarnation number before reincarnating a node by incrementing its incarnation number. This reduces the number of gossips being sent in the case of reincarnation. Which will result in faster memberlist convergence.

### DIFF
--- a/lib/membership/member.js
+++ b/lib/membership/member.js
@@ -178,6 +178,10 @@ Member.prototype._isLocalOverride = function _isLocalOverride(update) {
         return false;
     }
 
+    if (this.incarnationNumber > update.incarnationNumber) {
+        return false;
+    }
+
     return update.status === Member.Status.faulty ||
         update.status === Member.Status.suspect ||
         update.status === Member.Status.tombstone;
--- a/test/run-shared-integration-tests
+++ b/test/run-shared-integration-tests
@@ -6,7 +6,7 @@ set -eo pipefail
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="master"
+declare ringpop_common_branch="fix/double-reincarnation"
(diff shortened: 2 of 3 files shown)


test/run-shared-integration-tests
```diff
@@ -6,7 +6,7 @@ set -eo pipefail
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="master"
+declare ringpop_common_branch="fix/double-reincarnation"
```
mennopruijssers: Make sure to change this back before merging to master

mennopruijssers:
LGTM

lib/membership/member.js
```diff
@@ -178,6 +178,10 @@ Member.prototype._isLocalOverride = function _isLocalOverride(update) {
         return false;
     }
 
+    if (this.incarnationNumber > update.incarnationNumber) {
+        return false;
```
Motiejus: Can we create a unit test for this branch? Looking at `test/unit/member_test.js`, I think it should be quite easy.

thanodnl: Good point, I have added you (and the rest of the world) a test: https://github.com/uber/ringpop-node/pull/284/commits/f93890b4c0e5bb5e60b0778a5dc5ad9ab4c08e49

test/unit/membership_test.js
```diff
@@ -102,6 +102,29 @@ testRingpop('change that overrides the local status should be overwritten to a c
     assert.equals(member.status, Member.Status.alive, 'the status of the member should stay alive');
 });
 
+testRingpop('change that does not overrides the local status should not cause a reincarnation', function t(deps, assert) {
```
Motiejus: s/overrides/override/

Motiejus:
LGTM

PR #285, opened 2016-06-12
Fix typo

voidabhi:


Motiejus:
Sorry for the failed test, your change has nothing to do with it. We'll take a look at the test ASAP.

As for the actual proposed change, from the first sight it seems reasonable, but either me or someone from my team will get back to you later.

voidabhi:
np)

Motiejus:
Thank you for your contribution. We just checked with someone who wrote the code in the first place, and he things the tests are dormant and ought to be removed. See https://github.com/uber/ringpop-node/pull/286.

Closing.

Issue #287, opened 2016-06-17
npm install ringpop build error 


imjerryyang:
OS : MAC OS X EL
Node Version: v6.2.1, v4.4.5, v4.2.4, 0.12.9
Step to reproduce:
1 open terminal
2 npm install ringpop / npm install ringpop -save

```> toobusy@0.2.4 install /Users/xxxx/work/testing/node_modules/ringpop/node_modules/toobusy
> node-gyp rebuild

  CXX(target) Release/obj.target/toobusy/toobusy.o
../toobusy.cc:25:29: error: unknown type name 'Arguments'; did you mean 'v8::internal::Arguments'?
Handle<Value> TooBusy(const Arguments& args) {
                            ^~~~~~~~~
                            v8::internal::Arguments
/Users/xxxx/.node-gyp/4.2.4/include/node/v8.h:139:7: note: 'v8::internal::Arguments' declared here
class Arguments;
      ^
../toobusy.cc:37:20: error: no matching function for call to 'True'
    return block ? True() : False();
                   ^~~~
/Users/xxxx/.node-gyp/4.2.4/include/node/v8.h:8139:16: note: candidate function not viable: requires single argument 'isolate', but no arguments were provided
Local<Boolean> True(Isolate* isolate) {
               ^
../toobusy.cc:37:29: error: no matching function for call to 'False'
    return block ? True() : False();
                            ^~~~~
/Users/xxxx/.node-gyp/4.2.4/include/node/v8.h:8148:16: note: candidate function not viable: requires single argument 'isolate', but no arguments were provided
Local<Boolean> False(Isolate* isolate) {
               ^
../toobusy.cc:40:30: error: unknown type name 'Arguments'; did you mean 'v8::internal::Arguments'?
Handle<Value> ShutDown(const Arguments& args) {
                             ^~~~~~~~~
                             v8::internal::Arguments
/Users/xxxx/.node-gyp/4.2.4/include/node/v8.h:139:7: note: 'v8::internal::Arguments' declared here
class Arguments;
      ^
../toobusy.cc:45:12: error: no matching function for call to 'Undefined'
    return Undefined();
           ^~~~~~~~~
/Users/xxxx/.node-gyp/4.2.4/include/node/v8.h:315:27: note: candidate function not viable: requires single argument 'isolate', but no arguments were provided
  friend Local<Primitive> Undefined(Isolate* isolate);
                          ^
[...]

PR #288, opened 2016-06-17
Add requirements to README

dansimau:
Add a section to the README documenting the requirements for ringpop-node.

This should help people run into trouble when trying to `npm install` on a newer version of node[1].

[1]: https://github.com/uber/ringpop-node/issues/287

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ shard your application in a way that's scalable and fault tolerant.
 
 # Requirements
 
-* Node 0.10.x
+* Node 0.10 (0.10.32 or higher)
 
 # Installation
 To install Ringpop for usage as a library:
--- a/README.md
+++ b/README.md
@@ -4,6 +4,10 @@ applications. It maintains a consistent hash ring on top of a membership protoco
 and provides request forwarding as a routing convenience. It can be used to
 shard your application in a way that's scalable and fault tolerant.
 
+# Requirements
+
+* Node 0.10.x
+
 # Installation
 To install Ringpop for usage as a library:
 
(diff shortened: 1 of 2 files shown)


Motiejus:
LGTM

thanodnl:
The change to the readme looks good and can be merged, but shall we also specify the version in the `package.json` under the [`engines` section](https://docs.npmjs.com/files/package.json#engines)

mennopruijssers:
LGTM but let's add the `engines`-section as well :)

thanodnl:
I'm not completely sure if `0.10.x` is the right version. I was unable to do an install of the dependencies with the following versions:

 - 0.10.0
 - 0.10.1
 - 0.10.2
 - 0.10.3

All seem to have troubles with our package json dependencies section where the `^` is used to ask for compatible versions of packages. Versions that worked for me are the following:

- 0.10.26
- 0.10.32

Even though the problem is not with `node` but rather with `nvm` I wonder if we should specify a higher version of node to make sure that `nvm install` works. We can also figure out which version of `nvm` has support for `^` versions but I couldn't easily find when support was added. `npm@2.8.2` has support (shipped with `node@0.10.26`) and `npm@1.2.14` (shipped with `node@0.10.0`) does not have support.

dansimau:
Ah thanks @thanodnl for looking into this.

Should we specify `^0.10.32` then (that is the version we typically test with anyway)?

Maybe I should do a sanity check of versions above 0.10.32.

dansimau:
Travis tests run on v0.10.45 successfully so I'm just gonna merge this.

mennopruijssers:
LGTM!

dansimau:
Merged in commit ee72383.

PR #291, opened 2016-07-25
Status refactor 1: Refactor membership to use localMember as the source of truth

mennopruijssers:
Use a in-memory object to represent once local state. This makes the code path that updates the local state more robust. Instead of changing the state of the current node by updating it using updates, we just change the `localMember`. 
Whenever a gossip comes in changing the current state, we can always "counter-gossip" with the current state.

_Note: this is the first PR of a serie of refactors:_

- [x] Use a local member object to maintain local state:  #291 (this PR)
- [x] Use a common method for status updates: #292
- [x] Remove `LeaveUpdate`-object: #293
- [x] Remove "unsupported" state update methods: #294
- [x] Add a `populate{Subject,Source}` to the Change Object: #296
- [x] Add a `_reincarnate`-method to centralize reincarnation logic: #291 (this PR)

### DIFF
--- a/lib/membership/index.js
+++ b/lib/membership/index.js
@@ -44,9 +44,7 @@ function Membership(opts) {
     this.checksum = null;
     this.stashedUpdates = [];
     this.decayTimer = null;
-    this.localMember = new Member(this.ringpop, new Update(this.ringpop.whoami(), Date.now(), Member.Status.alive, null));
-    this.members.push(this.localMember);
-    this.membersByAddress[this.localMember.address] = this.localMember;
+    this.localMember = null;
 }
 
 util.inherits(Membership, EventEmitter);
@@ -217,14 +215,19 @@ Membership.prototype.makeLocalAlive = function makeLocalAlive(){
  * @param status The new status (@see Member.Status)
  */
 Membership.prototype.setLocalStatus = function setLocalStatus(status) {
-    if (status === Member.Status.leave) {
-        this.emit('event',
-            new MembershipEvents.LocalMemberLeaveEvent(this.localMember, this.localMember.status));
+    if (this.localMember) {
+        if (status === Member.Status.leave) {
+            this.emit('event',
+                new MembershipEvents.LocalMemberLeaveEvent(this.localMember, this.localMember.status));
+        }
+
+        this.localMember.status = status;
+        this.localMember.incarnationNumber = Date.now();
+    } else {
+        this.localMember = new Member(this.ringpop, new Update(this.ringpop.whoami(), Date.now(), status, null));
+        this.members.push(this.localMember);
+        this.membersByAddress[this.localMember.address] = this.localMember;
     }
-
-    this.localMember.status = status;
-    this.localMember.incarnationNumber = Date.now();
-
     this._postLocalUpdate();
 };
 
--- a/lib/membership/index.js
+++ b/lib/membership/index.js
@@ -201,10 +201,21 @@ Membership.prototype.isPingable = function isPingable(member) {
             Member.isStatusPingable(member.status);
 };
 
+/**
+ * Change the status of the local member to alive
+ * @see Membership#setLocalStatus
+ * @see Member.Status.alive
+ */
 Membership.prototype.makeLocalAlive = function makeLocalAlive(){
     this.setLocalStatus(Member.Status.alive);
 };
 
+/**
+ * Change the status of the local member. This will also bump the incarnation
+ * number of the local member.
+ *
+ * @param status The new status (@see Member.Status)
+ */
 Membership.prototype.setLocalStatus = function setLocalStatus(status) {
     if (status === Member.Status.leave) {
         this.emit('event',
@@ -217,6 +228,11 @@ Membership.prototype.setLocalStatus = function setLocalStatus(status) {
     this._postLocalUpdate();
 };
 
+/**
+ * Post an 'updated' event describing the local member's current state and
+ * recompute the membership checksum.
+ * @private
+ */
 Membership.prototype._postLocalUpdate = function _postLocalUpdate(){
     var update = new Update(this.localMember.address, this.localMember.incarnationNumber, this.localMember.status, this.localMember);
     this.computeChecksum();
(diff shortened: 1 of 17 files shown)


lib/membership/member.js
```diff
@@ -79,40 +79,10 @@ Member.prototype.decayDampScore = function decayDampScore() {
...
-        };
-    } else if (!this._isOtherOverride(update)) {
-        return;
-    }
-
-    // We've got an update. Apply all-the-things.
-    var oldStatus = this.status;
+// We've got an update. Apply all-the-things.
+Member.prototype.applyUpdate = function applyUpdate(update) {
     if (this.status !== update.status) {
```
thanodnl: I think we could remove the if statement for readability as well. Both with and without this check the value of `this.status` will be equal to `update.status`.

lib/membership/member.js
```diff
@@ -79,40 +79,10 @@ Member.prototype.decayDampScore = function decayDampScore() {
...
-
-        if (this.address === this.ringpop.whoami()) {
-            if (this.status === Member.Status.leave) {
-                this.ringpop.membership.emit('event',
-                    new events.LocalMemberLeaveEvent(this, oldStatus));
-            }
-        }
     }
 
     if (this.incarnationNumber !== update.incarnationNumber) {
```
thanodnl: same goes for this check. Yes, I favor small and concise code :)

lib/membership/member.js
mennopruijssers: Didn't want to change too much code but this make total sense.

test/unit/membership_test.js
```diff
@@ -204,7 +191,7 @@ testRingpop('cannot evict self', function t(deps, assert) {
 testRingpop('generate checksums string preserves order of members', function t(deps, assert) {
     var membership = deps.membership;
 
-    for (var i = 0; i < 100; i++) {
+    for (var i = 1; i < 100; i++) {
```
thanodnl: can you add a comment why we start? It took me some time to understand, and I hope I understand it correctly that this prevents to call `makeAlive` on the local member (?).

lib/membership/index.js
```diff
@@ -199,11 +199,66 @@ Membership.prototype.isPingable = function isPingable(member) {
...
+ * Bump the incarnation number of the local member and return an update that
+ * described the current state of the local member.
+ *
+ * @returns {Update} The update to gossip around
+ * @private
+ */
+Membership.prototype._reincarnate = function _reincarnate() {
+    this.localMember.incarnationNumber = Date.now();
+
+    return new Update(this.localMember.address, this.localMember.incarnationNumber, this.localMember.status, this.localMember);
```
thanodnl: With the changes in mind we want to make later does it make sense to refactor the `new Update(...)` call to take two members instead of all fields separated? We can also hold of to a later state when we actually extend the membership state with labels.

test/unit/discover_provider_healer_test.js
```diff
@@ -188,6 +188,7 @@ test('DiscoverProviderHeal.heal - only attempt to heal faulty (or worse) nodes',
         hostPort: '127.0.0.1:3000',
         discoverProviderHealerMaxFailures: maxFailures
     });
+    ringpop.membership.makeLocalAlive(ringpop.whoami(), Date.now());
```
thanodnl: the two arguments here are not used iirc.

test/unit/membership_test.js
mennopruijssers: correct. I'll add a comment
(side note: this change is in fact not necessary because calling makeAlive for the local member will just re-incarnate the local member again)

lib/membership/index.js
mennopruijssers: I was planning to do that in a separate PR (see the description of this PR). This one is already big enough ;)

lib/membership/member.js
```diff
@@ -250,4 +184,46 @@ Member.statusPrecedence = function statusPrecedence(status) {
...
+ * Determine if the incoming change should be processed or not.
+ * The update should be processed if:
+ *  - the update is about an unknown member and the status is not 'tombstone'
+ *  - the new status takes precedence over the current known status
+ *    (@see Member~statusPrecedence)
+ *  - the incarnation number is newer
+
+ * @param {Member} [member] The current member or null when the member is currently
+ *  unknown.
+ * @param {object} change The incoming update
```
thanodnl: can we use a more precise definition for a change than just a plane object? Wasn't there an `IMember` or will this be added in a later PR?

thanodnl:
LGTM

lib/membership/member.js
mennopruijssers: Good point, I added it to #296 (the one where we define `IMember`)

lib/membership/index.js
```diff
@@ -199,11 +199,66 @@ Membership.prototype.isPingable = function isPingable(member) {
...
+
+/**
+ * Bump the incarnation number of the local member and return an update that
+ * described the current state of the local member.
+ *
+ * @returns {Update} The update to gossip around
+ * @private
+ */
+Membership.prototype._reincarnate = function _reincarnate() {
+    this.localMember.incarnationNumber = Date.now();
```
Motiejus: Not sure if you are doing this in the later PRs, but maybe we should make it like this?

```javascript
this.localMember.incarnationNumber = max(this.localMember.incarnationNumber + 1, Date.now());
```

Motiejus:
Besides the [incarnation number change][1] (non-blocking, just let me know what you think), LGTM.

[1]: https://github.com/uber/ringpop-node/pull/291#discussion_r73132206

lib/membership/index.js
mennopruijssers: I discussed this with nils and we decided to not do it right now. These PR's are refactoring stuff; not changing behaviour. If we want to move forward with the `max(inc + 1, Date.now()` we can do it at a later stage; we probably have to double check if there's not anything else breaking. 

 /cc @thanodnl

mennopruijssers:
As discussed offline with @thanodnl I renamed `shouldProcessChange` to `shouldProcessGossip`. He's doing the same in ringpop-go so we have them in sync.

PR #294, opened 2016-07-25
Status refactor 4: Remove makeLeave and makeAlive

mennopruijssers:
Since it's not allowed by the SWIM protocol for a node to change a member's status to `alive` or `leave`,  let's remove these methods. Tests can use the generic `makeChange`-function.


thanodnl:
LGTM (LGTM is in the `ALT`)

Motiejus:
👍

PR #296, opened 2016-07-28
Status refactor 5: Change update constructor

mennopruijssers:
Change the update-constructor to take a subject and source.
To be able to document everything correctly, I had to:
- add `@enum` to `Member.Status`,
- Add an interface for member (`IMember`)

### DIFF
--- a/lib/membership/index.js
+++ b/lib/membership/index.js
@@ -217,14 +217,14 @@ Membership.prototype.makeLocalAlive = function makeLocalAlive(){
 Membership.prototype._reincarnate = function _reincarnate() {
     this.localMember.incarnationNumber = Date.now();
 
-    return new Update(this.localMember.address, this.localMember.incarnationNumber, this.localMember.status, this.localMember);
+    return new Update(this.localMember, this.localMember);
 };
 
 /**
  * Change the status of the local member. This will also bump the incarnation
  * number of the local member.
  *
- * @param status The new status (@see Member.Status)
+ * @param {Member.Status} status The new status
  */
 Membership.prototype.setLocalStatus = function setLocalStatus(status) {
     if (this.localMember) {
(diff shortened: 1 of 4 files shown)


lib/membership/update.js
```diff
@@ -24,15 +24,29 @@ var Member = require('./member.js');
...
-    this.incarnationNumber = incarnationNumber;
-    this.status = status;
+/**
+ * Create a new Update
+ * @param {IMember} subject the subject of the update.
+ * @param {IMember} [source] the source of the update.
+ * @constructor
+ */
+function Update(subject, source) {
+    if (!(this instanceof Update)) {
```
thanodnl: This branch is new in this PR, is it used or shall we remove it since you also indicated via `@constructor` that it should be invoked with `new`. Especially since the `Update` is probably not called as a function in the rest of the code.

thanodnl:
Only 1 minor question about the defensive programming in the `Update` constructor, but overall this PR __LGTM__.

lib/membership/update.js
mennopruijssers: It's not used but most other objects in the repo are using a `instanceof`-check to make sure it's invoked using `new`. When cleaning up this class I thought it made sense to add it here as well but if you prefer I can remove it.

thanodnl: If it is used in most of the objects here we can keep it this way, it will make the code more consistent. I was just pointing out that it got added here and not used. But consistency will win in my book in the end.

👍

lib/membership/index.js
```diff
@@ -235,7 +235,11 @@ Membership.prototype.setLocalStatus = function setLocalStatus(status) {
 
         this.localMember.status = status;
     } else {
-        this.localMember = new Member(this.ringpop, new Update(this.ringpop.whoami(), Date.now(), status, null));
+        this.localMember = new Member(this.ringpop, new Update({
```
Motiejus: 👍

Motiejus:
well, yay! 👍

PR #299, opened 2016-08-05
[self-evict 1] Support self eviction as a way of graceful shutdown

mennopruijssers:
This PR adds a way to add self-eviction: marking the local member faulty before shutdown. 
It adds two functions to ringpop's public API:
- `registerSelfEvictHooks`: Provide a way to register hooks
- `selfEvict`: initiate the self eviction process

TODO in seperate PR(s): 
- [x] ping members before ending the sequence so the self-eviction is guaranteed to be gossiped (#300)

### DIFF
--- a/lib/self-evict.js
+++ b/lib/self-evict.js
@@ -1,0 +1,319 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+var _ = require('underscore');
+var TypedError = require('error/typed');
+
+var errors = require('./errors');
+var Member = require('./membership/member');
+
+
+/**
+ * @typedef {Object} Phase
+ * @property {SelfEvict.PhaseNames} phase the name of the phase.
+ * @property {number} ts the unix timestamp in milliseconds when the phase started.
+ * @property {number} [duration] the duration in milliseconds of the phase.
+ */
+
+/**
+ * This object defines the hooks. At least one of hook-methods (i.e.
+ * preEvict or postEvict, or both) should be defined.
+ * @typedef {Object} SelfEvictHooks
+ * @property {string} name The name of the hook
+ * @property {SelfEvict~Hook} [preEvict] The preEvict hook to register.
+ * @property {SelfEvict~Hook} [postEvict] The postEvict hook to register.
+ */
+/**
+ * A pre or post self eviction hook.
+ * @callback SelfEvict~Hook
+ * @param {SelfEvict~HookCallback} cb the callback to be called when the hook completes
+ */
+
+/**
+ * A callback used when calling Hooks
+ * @callback SelfEvict~HookCallback
+ */
+
+/**
+ * Self eviction provides a way to graceful shut down a ringpop service.
+ * A self eviction consists of the following phases:
+ * 1. PreEvict: the preEvict hooks are invoked;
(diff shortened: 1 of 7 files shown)


lib/self-evict.js
```diff
@@ -0,0 +1,303 @@
...
+    var self = this;
+    this._runHooks(this.preEvictHooks, function hooksDone() {
+        process.nextTick(self._evict.bind(self));
+    });
+};
+
+SelfEvict.prototype._evict = function _evict() {
+    this._transitionTo(PhaseNames.Evicting);
+
+    this.ringpop.membership.setLocalStatus(Member.Status.faulty);
```
thanodnl: I think we should forcefully ping nodes here. If there are no posthooks or they take a really short time our faulty declaration might not reach any node before shutting down. This will cause the normal suspect period to kick in.

As discussed the number and/or targets to ping should be configurable. You might want to ping all the new owners of part of your keyspace. You might also decide to quickly send 3 pings to get the information into the network before disappearing.

Only after you have confirmation that the nodes have received the gossip it is safe to transition to the postEviction state I believe.

mennopruijssers: As mentioned in the description: I'll add that in a separate PR to keep this one smaller/less complicated.

lib/self-evict.js
```diff
@@ -0,0 +1,303 @@
...
+var _ = require('underscore');
+var TypedError = require('error/typed');
+
+var errors = require('./errors');
+var Member = require('./membership/member');
+
+
+/**
+ * @typedef {Object} Phase
+ * @property {SelfEvict.PhaseNames} phase the phase
```
Motiejus: Can you clarify the description of the variable? `phase the phase` is not very meaningful to me :(

lib/self-evict.js
```diff
@@ -0,0 +1,303 @@
...
+ * @property {number} ts the timestamp when the phase started
```
Motiejus: Can you clarify units, e.g. "unix timestamp in milliseconds".

lib/self-evict.js
```diff
@@ -0,0 +1,303 @@
...
+ * @property {number} [duration] the duration of a phase
```
Motiejus: Units please. :-)

lib/self-evict.js
```diff
@@ -0,0 +1,303 @@
...
+ */
+
+/**
+ * A callback used when calling Hooks
+ * @callback SelfEvict~HookCallback
+ */
+
+/**
+ * Self eviction provides a way to graceful shut down a ringpop service.
+ * A self eviction  consists of the following phases:
```
Motiejus: Double space

lib/self-evict.js
```diff
@@ -0,0 +1,303 @@
...
+ * 1. PreEvict: the preEvict hooks are invoked;
+ * 2. Evict: The node will mark  itself as faulty causing an eviction out of the membership;
+ * 3. PostEvict: the postEvict hooks are invoke;
```
Motiejus: s/invoke/invoked

lib/self-evict.js
```diff
@@ -0,0 +1,303 @@
...
+     */
+    this.phases = [];
+}
+
+/**
+ * Get the current  phase or null when self eviction isn't initiated
+ * @return {?Phase}
+ */
+SelfEvict.prototype.currentPhase = function currentPhase() {
+    if (this.phases && this.phases.length > 0) {
```
Motiejus: Should we be that defensive here? `this.phases` is initialized in the constructor to be an array...

lib/self-evict.js
```diff
@@ -0,0 +1,303 @@
...
+        return this.phases[this.phases.length - 1];
+    }
+    return null;
+};
+
+/**
+ * Register hooks in the eviction sequence. At least one of hook-methods should
+ * be defined.
```
Motiejus: Maybe clarify like this:

At least one of the hook-methods (i.e. either `preEvict`, or `postEvict`, or both) should be defined in a single invocation.

Makes it easier to parse.

Motiejus: You could also document that the hooks are invoked in an undefined order

lib/self-evict.js
```diff
@@ -0,0 +1,303 @@
...
+
+    var self = this;
+    var hookTimings = {};
+
+    var afterHook = _.after(names.length, function afterHooks() {
+        logger.debug('ringpop self eviction done running hooks', {
+            local: self.ringpop.whoami(),
+            totalDuration: Date.now() - start,
+            hookDurations: hookTimings,
+            phase: self.currentPhase()
```
Motiejus: Maybe it would be useful to log the order in which the hooks were invoked?

Motiejus:
My comments are non-blocking, LGTM after you address them.

I am not sure about the order of hooks myself (apart from documenting it), maybe it's not useful at all. I am not very sure in how we intend developers to use them.

thanodnl:
LGTM

lib/self-evict.js
```diff
@@ -131,11 +131,11 @@ SelfEvict.prototype.registerHooks = function registerHooks(hooks) {
     }
 
     if (hooks.preEvict) {
-        this.preEvictHooks[name] = hooks.preEvict;
+        this.preEvictHooks[name] = hooks.preEvict.bind(hooks);
```
mennopruijssers: The other option is to not have two separate maps but one (e.g. `hooks`). However, that would require in the way they are ran. Right now `_runHooks` is passed in an object with all the hooks to run; that have to be a `type` (`preEvict` or `postEvict`). I than have to loop through the full `this.hooks`-object, check if `hook[type]` is available, and invoke it when it is. What would you two prefer?

lib/self-evict.js
```diff
@@ -0,0 +1,319 @@
...
+
+    /**
+     * The history of phases
+     * @type {[Phase]}
+     */
+    this.phases = [];
+}
+
+/**
+ * Get the current  phase or null when self eviction isn't initiated
```
thanodnl: nit: double space after 'current'

Motiejus:
LGTM

PR #300, opened 2016-08-08
[self-evict 2] Gossip self eviction

mennopruijssers:
Ping other members on self-eviction.

This PR is dependent on #299

### DIFF
--- a/lib/self-evict.js
+++ b/lib/self-evict.js
@@ -186,14 +186,13 @@ SelfEvict.prototype._evict = function _evict() {
 
     var pingEnabled = this.ringpop.config.get('selfEvictionPingEnabled');
 
-    var restartGossip = false;
-    // Stop gossip when currently gossiping and ping is enabled
-    if (!this.ringpop.gossip.isStopped && pingEnabled) {
-        this.ringpop.gossip.stop();
-        restartGossip = true;
-    }
-
     if (pingEnabled) {
+        var restartGossip = false;
+        // Stop gossip when currently gossiping and ping is enabled
+        if (!this.ringpop.gossip.isStopped) {
+            this.ringpop.gossip.stop();
+            restartGossip = true;
+        }
         waitUntilNotPinging();
 
         function waitUntilNotPinging() {
@@ -226,20 +225,19 @@ SelfEvict.prototype._evict = function _evict() {
                     numberOfPings--;
                     self.ringpop.gossip.tick(tick);
                 } else {
-                    done();
+                    if (restartGossip) {
+                        self.ringpop.gossip.start();
+                    }
+                    evictDone();
                 }
             }
         }
     } else {
         self.ringpop.membership.setLocalStatus(Member.Status.faulty);
-        done();
+        evictDone();
     }
 
-    function done() {
-        if (restartGossip) {
-            // enable pinging again
-            self.ringpop.gossip.start();
-        }
+    function evictDone() {
         process.nextTick(self._postEvict.bind(self));
     }
 };
--- a/lib/self-evict.js
+++ b/lib/self-evict.js
@@ -23,6 +23,7 @@ var TypedError = require('error/typed');
 
 var errors = require('./errors');
 var Member = require('./membership/member');
+var sendPing = require('./gossip/ping-sender.js').sendPing;
 
 
 /**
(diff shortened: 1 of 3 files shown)


test/unit/self-evict-test.js
```diff
@@ -168,6 +178,76 @@ testRingpop({async: true}, 'self evict sequence invokes hooks', function t(deps,
...
+
+    ringpop.gossip.tick = function tick(cb) {
+        assert.ok('gossip ticked!');
+        cb();
+    };
+    var selfEvict = new SelfEvict(ringpop);
+    selfEvict.initiate(cleanup);
+});
+
+testRingpop({async: true}, 'self evict does not ticks membership when "selfEvictionPingEnabled" is disabled', function t(deps, assert, cleanup) {
```
Motiejus: s/ticks/tick/

Motiejus:
LGTM

lib/self-evict.js
```diff
@@ -179,10 +180,68 @@ SelfEvict.prototype._preEvict = function _preEvict() {
...
+
     this._transitionTo(PhaseNames.Evicting);
 
-    this.ringpop.membership.setLocalStatus(Member.Status.faulty);
-    process.nextTick(this._postEvict.bind(this));
+    var pingEnabled = this.ringpop.config.get('selfEvictionPingEnabled');
+
+    var restartGossip = false;
+    // Stop gossip when currently gossiping and ping is enabled
+    if (!this.ringpop.gossip.isStopped && pingEnabled) {
```
thanodnl: I think the code is a bit more readable when you remove the `pingEnabled` check here and move the complete block in the `pingEnabled` condition block below.

Motiejus:
Should we put a section about self-eviction to the [architecture page in public docs][1]?

[1]: http://ringpop.readthedocs.io/en/latest/architecture_design.html#how-ringpop-works

lib/self-evict.js
```diff
@@ -179,10 +180,68 @@ SelfEvict.prototype._preEvict = function _preEvict() {
...
+                self.ringpop.dissemination.maxPiggybackCount,
+                maxNumberOfTicks
+            );
+
+            tick();
+
+            function tick() {
+                if (numberOfPings > 0) {
+                    numberOfPings--;
+                    self.ringpop.gossip.tick(tick);
```
thanodnl: `gossip.tick()` executes a `ping` and when it fails it fans out a `pingReq`, I wonder if we want to wait for ping timeouts and then a pingreq which involves a timeout as well. I think only doing a ping is enough for the shutting down phase.

thanodnl:
I miss a parallelism factor to control how many concurrent pings we will be sending during self eviction. Is that still sth we want to have in this feature or are willing to take the hit on shutdown speed if you are pinging nodes that just went away? Same about my comment to use `gossip.tick`,  are we ok with the slower shutdown phase by including pingreqs in the process?

lib/self-evict.js
mennopruijssers: Agreed; however there's currently not an api exposed to ping a member (or multiple members) without fan-out on `gossip`. 
Would you prefer adding an optional `opts` argument to `tick` (e.g. `tick({fanOut: false}, cb)`) or do you have a better idea?

mennopruijssers:
@thanodnl I wanted to start simple without parallelism and test out the impact on shutdown speed. If shutdown is taking too long we can improve by pinging in parallel. This would however, require some changes in the way the `tick` function is currently implemented: currently it allows only one `ping` running at a time.

mennopruijssers:
@Motiejus probably yes but lets first try to stabilize the feature :)

thanodnl:
I think the easiest is to not rely on the tick feature but extract the part that sends the ping (which is already quite abstract with `sendPing`) into a reusable component and use that one instead. That would solve the `pingReq` concerns at the same time. But we can start with this and indeed measure how fast a shutdown is before optimizing.

mennopruijssers:
> I think the easiest is to not rely on the tick feature but extract the part that sends the ping (which is already quite abstract with `sendPing`) into a reusable component and use that one instead. That would solve the `pingReq` concerns at the same time. But we can start with this and indeed measure how fast a shutdown is before optimizing.
I changed to code to use `sendPing`. Makes sense, thanks!

Motiejus:
👍

lib/self-evict.js
```diff
@@ -182,65 +183,53 @@ SelfEvict.prototype._preEvict = function _preEvict() {
...
+        var maxNumberOfTicks = Math.ceil(numberOfPingableMembers * selfEvictionMaxPingRatio);
 
-            function tick() {
-                if (numberOfPings > 0) {
-                    numberOfPings--;
-                    self.ringpop.gossip.tick(tick);
-                } else {
-                    done();
-                }
+        // Never ping more than the maximum piggy back counter.
```
thanodnl: s/piggy back/piggyback

lib/self-evict.js
```diff
@@ -182,65 +183,53 @@ SelfEvict.prototype._preEvict = function _preEvict() {
...
+        var numberOfPings = Math.min(
```
thanodnl: lets also add `numberOfPingableMembers` to the min. There are cases where `numberOfPings` might exceed the number of pingable members.

First the `selfEvictionMaxPingRatio` must be bigger then 1
Secondly the `maxPiggybackCount` needs to exceed the number of members. This is true on cluster sizes less then 30 I believe.

test/unit/self-evict-test.js
```diff
@@ -184,68 +184,72 @@ testRingpop({async: true}, 'self evict ticks membership to speed up gossip', fun
...
+        protocolPing: function(opts, body, cb) {
+            var target = opts.host;
+            if (target === '127.0.0.1:30002') {
+                assert.ok('127.0.0.1:30002 is pinged!');
+                cb(null, {changes: []});
+            } else if (target === '127.0.0.1:30003') {
+                assert.ok('127.0.0.1:30003 is pinged!');
+                cb('error', null);
+            } else {
+                assert.fail('incorrect target');
```
thanodnl: Is this assertion not being called? I would expect it to be called with an empty target base on my earlier comment.

lib/self-evict.js
mennopruijssers: Good catch!

test/unit/self-evict-test.js
mennopruijssers: No: it shouldn't be called because the third member is not pingable. The assert makes sure it's not called.

thanodnl:
LGTM

lib/self-evict.js
```diff
@@ -177,10 +179,57 @@ SelfEvict.prototype._preEvict = function _preEvict() {
...
+    if (pingEnabled) {
+        var selfEvictionMaxPingRatio = ringpop.config.get('selfEvictionMaxPingRatio');
+        var membership = ringpop.membership;
+
+        var pingableMembers = _.filter(membership.members, membership.isPingable.bind(membership));
+        var numberOfPingableMembers = pingableMembers.length;
+        var maxNumberOfPings = Math.ceil(numberOfPingableMembers * selfEvictionMaxPingRatio);
+
+        // Never ping more than the maximum piggyback counter and/or number of pingable members.
+        var numberOfPings = Math.min(
```
thanodnl: I found it hard to read through this block of code. A lot is happening, for both peer selection and calculation of how many we should ping. For readability and testability I would love to find a way to pull this logics into its own function.

I think we can do so by returning the list of all candidates that will be pinged and use `list.length` when we need to get the number of nodes we are planning to ping.

lib/self-evict.js
```diff
@@ -178,6 +178,24 @@ SelfEvict.prototype._preEvict = function _preEvict() {
     });
 };
 
+SelfEvict.prototype._getMembersToPing = function(ringpop) {
```
thanodnl: Nit: I think `ringpop` can be obtained from the object via `this.ringpop`. What is the benefit of passing it in?

mennopruijssers: :facepalm:

Motiejus:
LGTM

PR #301, opened 2016-08-09
[self-evict 3] self evict on shutdown in testpop

mennopruijssers:
Let testpop `selfEvict` on `ctrl+c` (`SIGINT`).

### DIFF
--- a/main.js
+++ b/main.js
@@ -92,13 +92,18 @@ function main(args) {
 
     ringpop.setupChannel();
 
-    var shuttingDown = false;
-    process.on('SIGTERM', signalHandler);
-    process.on('SIGINT', signalHandler);
+    process.once('SIGTERM', signalHandler(false));
+    process.once('SIGINT', signalHandler(true));
 
-    function signalHandler() {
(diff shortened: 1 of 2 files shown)


main.js
```diff
@@ -92,6 +92,22 @@ function main(args) {
...
+            ringpop.selfEvict(function afterSelfEvict(err){
+                if (err) {
+                    console.error('Failure during selfEvict: ' + err);
+                    process.exit(1);
+                    return;
+                }
+                process.exit(0);
+            });
+        } else {
+            process.exit(1);
```
thanodnl: I think we should not exit here, or we should log a message that we received a second interrupt and therefore are aborting whatever we are doing and shut down.

What is the reasoning behing the emergency exit here?

mennopruijssers: Force kill by hitting `ctrl+c` twice.

mennopruijssers:
I added better double interrupt handling and console logging.

thanodnl:
LGTM

Motiejus:
LGTM

PR #302, opened 2016-08-10
Disable healing on self

mennopruijssers:
Since we now support a self declared faulty on the local member, during partition healing we need to explicitly exclude ourself from the list of targets during partition healing.
previously this was implied because the local member always changed it's status to `alive`.


thanodnl:
LGTM

CorgiMan:
Looks good!

PR #304, opened 2016-09-07
Add node4 support

Raynos:
This PR has been tested in the production rt-supply service.

This PR contains all the work by @sashahilton00

r: @thanodnl @Motiejus

### DIFF
--- a/test/run-shared-integration-tests
+++ b/test/run-shared-integration-tests
@@ -6,7 +6,7 @@ set -eo pipefail
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="menno/node-4-compat"
+declare ringpop_common_branch="master"
 declare tap_filter="${ringpop_common_dir}/test/tap-filter"
 
 declare test_cluster_sizes="1 2 3 4 5 10"
--- a/test/run-shared-integration-tests
+++ b/test/run-shared-integration-tests
@@ -6,7 +6,7 @@ set -eo pipefail
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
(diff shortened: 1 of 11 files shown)


test/run-shared-integration-tests
```diff
@@ -6,7 +6,7 @@ set -eo pipefail
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="master"
+declare ringpop_common_branch="menno/node-4-compat"
```
Motiejus: Is that branch really necessary? Looks like the only change is tchannel:

```
motiejus ~/Uber/ringpop-common $ git diff $(git merge-base origin/master origin/menno/node-4-compat) origin/menno/node-4-compat
diff --git a/test/package.json b/test/package.json
index bdb328d..05288d2 100644
--- a/test/package.json
+++ b/test/package.json
@@ -20,7 +20,7 @@
     "tap-parser": "^1.2.2",
     "tape": "^4.2.0",
     "tape-catch": "^1.0.5",
-    "tchannel": "^2.4.6",
+    "tchannel": "^3.6.13",
     "underscore": "^1.8.3"
   }
 }
```

Should we update the tchannel on master in integration tests instead?

Motiejus:
Except for the comment about the branch of integration tests, LGTM.

sashahilton00:
I'd update master for choice and stop using the branch. Your call though.

ianp:
Would it be a good idea to also add 6 to the version list in `.travis.yml`?

Raynos:
Addressed feedback, merging PR.

Raynos:
If someone could cut a new patch version that would be great.

thanodnl:
I have validated the compatibility of the farmhash between the new version and older versions to make sure the upgrade path it simple. We will make a new release later this week and publish it to npm.

PR #306, opened 2016-09-23
Update version of hammock for perf boost

Raynos:
The hammock library is used in the ringpop proxy code
and had some performance issues in response buffering.

These are fixed in the latest version of hammock

r: @thanodl @motiejus @mennopruijssers

### DIFF
--- a/test/integration/proxy_test.js
+++ b/test/integration/proxy_test.js
@@ -715,7 +715,7 @@ test('reroutes retry to remote', function t(assert) {
             retrySchedule: [1]
         }, function onRequest(err, res) {
             assert.ifError(err, 'no error occurs');
-            assert.equal(res.body, 'rerouted', 'response from rerouted request is correct');
+            assert.equal(res.body.toString(), 'rerouted', 'response from rerouted request is correct');
 
             cluster.destroy();
             assert.end();
@@ -896,7 +896,7 @@ test('can serialize response body', function t(assert) {
             assert.ifError(err);
 
             assert.equal(resp.statusCode, 200);
-            assert.equal(resp.body, 'hello');
+            assert.equal(resp.body.toString(), 'hello');
 
             cluster.destroy();
             assert.end();
--- a/test/integration/proxy_test.js
+++ b/test/integration/proxy_test.js
@@ -715,7 +715,11 @@ test('reroutes retry to remote', function t(assert) {
             retrySchedule: [1]
         }, function onRequest(err, res) {
             assert.ifError(err, 'no error occurs');
-            assert.equal(res.body.toString(), 'rerouted', 'response from rerouted request is correct');
+            assert.equal(
+                res.body.toString('hex'),
+                new Buffer('rerouted').toString('hex'),
+                'response from rerouted request is correct'
+            );
 
             cluster.destroy();
             assert.end();
@@ -896,7 +900,10 @@ test('can serialize response body', function t(assert) {
             assert.ifError(err);
 
             assert.equal(resp.statusCode, 200);
-            assert.equal(resp.body.toString(), 'hello');
+            assert.equal(
+                resp.body.toString('hex'),
+                new Buffer('hello').toString('hex')
+            );
 
             cluster.destroy();
             assert.end();
--- a/index.js
+++ b/index.js
@@ -691,6 +691,13 @@ RingPop.prototype.handleOrProxyAll =
                 keys: keysByDest[dest]
             });
             if ((--pending === 0 || err) && cb) {
+                for (var i = 0; i < responses.length; i++) {
+                    var r = responses[i];
+                    if (Buffer.isBuffer(r.body)) {
+                        r.body = r.body.toString('utf8');
+                    }
+                }
+
                 cb(err, responses);
                 cb = null;
             }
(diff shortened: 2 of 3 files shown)


Motiejus:
Hi Jake,

thank you for your pull request, and sorry for a late reply. Me and @CorgiMan will take a good look at this tomorrow AMS time.

Motiejus:
LGTM from me. I'll ping someone else in the team tomorrow to get another approval and will see through merging this.

Thank you for the contribution!

CorgiMan:
I guess we need to add ".toString()" to the tests to get them fixed. How can we be sure that this change doesn't affect our customers like it affected this test?

test/integration/proxy_test.js
```diff
@@ -715,7 +715,7 @@ test('reroutes retry to remote', function t(assert) {
             retrySchedule: [1]
         }, function onRequest(err, res) {
             assert.ifError(err, 'no error occurs');
-            assert.equal(res.body, 'rerouted', 'response from rerouted request is correct');
+            assert.equal(res.body.toString(), 'rerouted', 'response from rerouted request is correct');
```
Motiejus: I agree. If the API did not change, we shouldn't change the test. Can you help me understand why this test changed?

Raynos: The API changed to not convert Buffers to strings automatically.

The `res.body` is a `Buffer`, the performance optimization is not to convert it to a string.

I can do `toString('utf8')` to make it more clear, Or I can do: 

```
assert.equal(res.body.toString('hex'), new Buffer('rerouted').toString('hex'))
```

Raynos:
The performance optimization is that `res.body` used to be `String`, now it's `String | Buffer`.

The `hammock` library will do as little conversion between `String` and `Buffer` as it can.

This means we can now write a `Buffer` to `MockResponse`, and send that same `Buffer` instance through tchannel down the wire.

Previously we had `Buffer` -> `MockResponse` -> `String` -> `TChannel` -> `Buffer` -> `Socket`. We had converted from a Buffer -> String, only to convert the string straight back to a new Buffer before writing it o the socket.

CorgiMan:
It's not clear to me that this won't cause a degradation in ringpop's API. Can you elaborate why it wouldn't please? Thanks!

Raynos:
There are two places where `hammock.Response` is used.

The `handleOrProxy` common case is handled internally ( https://github.com/uber/ringpop-node/blob/master/lib/request-proxy/index.js#L242-L263 ).

Here we pass `resp.body` to tchannel which already supports `String | Buffer`.

The second place is `handleOrProxyAll` ( https://github.com/uber/ringpop-node/blob/master/index.js#L687-L697 )

Here we do return `cb(null, Array<MockResponse>)` so we do degrade the public interface of ringpop for the `handleOrProxyAll` method.

Raynos:
I've audited three services, `supply`, `demand`, `optic`.

Only `demand` uses `handleOrProxyAll`, It's current use cases of this method call all ignore the `response` array except for logging it. Changing the interface from `Array<{ body: String, ... }>` => `Array<{ body: String | Buffer, ... }>` will not have much impact.

I think the best option is to add a for loop in the implementation of `handleOrProxyAll` where we convert buffers back to strings for backwards compatibility.

This allows us to keep the `handleOrProxy` method using the optimized code path.

Raynos:
Great observation, thanks @CorgiMan

test/integration/proxy_test.js
```diff
@@ -715,7 +715,11 @@ test('reroutes retry to remote', function t(assert) {
             retrySchedule: [1]
         }, function onRequest(err, res) {
             assert.ifError(err, 'no error occurs');
-            assert.equal(res.body, 'rerouted', 'response from rerouted request is correct');
+            assert.equal(
+                res.body.toString('hex'),
+                new Buffer('rerouted').toString('hex'),
+                'response from rerouted request is correct'
```
Motiejus: If my understanding of the backwards-compatibility fix (604350879c41b823eb69eb04a1a7382d8ae961c3) is correct, we no longer need to change the tests, right? Body will be string at this point?

Motiejus:
Thank you for implementing the backwards-compatibility change. If my understanding is correct and we can leave the test unchanged, LGTM.

CorgiMan:
I agree with Motiejus. If we leave the test as it was you can land it!

Thanks for the investigation @Raynos

test/integration/proxy_test.js
Raynos: This test still breaks because its a unit test using an hammock. Basically `hammock` is used both in ringpop proxying and in the tests themself.

Since the `res` instance is allocated inside the test cluster we need to change the test.

Raynos:
The test still needs to change, beyond the two usages of `hammock` in ringpop, There are further uses of `hammock` in the test suite themselves.

The `res` instance in the test assertion is actually not allocated by ringpop, its allocated by the test. The `hammock` API used in the test suite has changed.

CorgiMan:
LGTM!

Raynos:
@Motiejus @thanodnl @CorgiMan 

Can we release a new patch version with this performance optimization ?

thanodnl:
@Raynos A patch release has been released an published on npmjs.

PR #307, opened 2016-10-18
Fix for self-eviction hooks

mennopruijssers:
After merging #299, #300 and #301, a test in ringpop-serial broke (probably bad merge); this fixes it.
Also added some extra type-checking on the hooks (`isFunction`).

### DIFF
--- a/lib/self-evict.js
+++ b/lib/self-evict.js
@@ -294,7 +294,7 @@ SelfEvict.prototype._runHooks = function _runHooks(type, done) {
     var ringpop = self.ringpop;
 
     var names = _.filter(_.keys(this.hooks), function filter(name) {
-        return _.has(self.hooks[name], type);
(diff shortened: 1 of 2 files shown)


lib/self-evict.js
```diff
@@ -294,7 +294,7 @@ SelfEvict.prototype._runHooks = function _runHooks(type, done) {
     var ringpop = self.ringpop;
 
     var names = _.filter(_.keys(this.hooks), function filter(name) {
-        return _.has(self.hooks[name], type);
+        return self.hooks[name][type];
```
thanodnl: This needs a bit of explanation. These lines seem to do the same.

Motiejus:
LGTM

lib/self-evict.js
thanodnl: Yay, good find.

PR #310, opened 2016-11-15
[Feature] Key consistent forwarding

thanodnl:
Currently ringpop can be configured to compare the hashring when a request is forwarded and drop the request if the hashrings are inconsistent. This causes a lot of requests that are actually forwarded to the correct node to fail, resulting in a lot of timeout and errors on rolling restarts.

The PR will add the possibility to loosen the requirement from ring consistency to key consistency eg. if the key from the forwarded request is believed to be owned by the receiving node based on a hashring lookup it will execute the request. When this happen both the forwarder and the receiver of the forwarded request agree on the key ownership for the key to the receiving end.

This showed a huge improvement in performance of a ringpop application during a rolling restart.

### DIFF
--- a/lib/request-proxy/index.js
+++ b/lib/request-proxy/index.js
@@ -44,6 +44,14 @@ var InvalidCheckSumError = TypedError({
     actual: null,
     expected: null
 });
+var InvalidKeyError = TypedError({
+    type: 'ringpop.request-proxy.invalid-key',
+    message: 'Expected the local node to resolve to the forwarded key.\n' +
+             'A lookup for "{key}" resulted in {target} instead of {whoami}.',
+    key: null,
+    target: null,
+    whoami: null
+});
 var TooManyRequestsError = TypedError({
     type: 'ringpop.request-proxy.too-many-requests',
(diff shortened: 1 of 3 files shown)


thanodnl:
Tests look good as well

test/unit/request_proxy_test.js
```diff
@@ -32,10 +32,13 @@ function createRingpop() {
     });
 }
 
-function createRequestProxy() {
-    return new RequestProxy({
-        ringpop: createRingpop()
-    });
+function createRequestProxy(opts) {
```
danielheller: Nice tests!

lib/request-proxy/index.js
```diff
@@ -221,6 +235,38 @@ RequestProxy.prototype.handleRequest = function handleRequest(head, body, cb) {
         }
     }
 
+    var whoami = ringpop.whoami();
```
thanodnl: Pulling this block to its own function will over complicate the flow since there is a pre-emptive escape in the block which is essential.

I will add a comment above this block (and for that matter to the block above as well) explaining why this is executed in place and always, eg. it gives stats on how many requests are executed on wrong hosts if when consistency is not enforced.

PR #313, opened 2016-12-20
Fix: Initial transition timers

mennopruijssers:
This PR fixes an issue where transition timers (`faulty` -> `tombstone`) were not scheduled correctly if the membership during a `join`-response contains a `faulty`-member. Instead of duplicating the code that schedules the timers, a method is extracted from the gossip handler and used on the `set`-event (emitted after the initial join).

### DIFF
--- a/test/run-shared-integration-tests
+++ b/test/run-shared-integration-tests
@@ -6,7 +6,7 @@ set -eo pipefail
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="fix/test-state-transistions"
+declare ringpop_common_branch="master"
 declare tap_filter="${ringpop_common_dir}/test/tap-filter"
 
 declare test_cluster_sizes="1 2 3 4 5 10"
--- a/test/run-shared-integration-tests
+++ b/test/run-shared-integration-tests
@@ -6,7 +6,7 @@ set -eo pipefail
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="master"
+declare ringpop_common_branch="fix/test-state-transistions"
 declare tap_filter="${ringpop_common_dir}/test/tap-filter"
 
 declare test_cluster_sizes="1 2 3 4 5 10"
--- a/test/unit/membership_test.js
+++ b/test/unit/membership_test.js
@@ -266,7 +266,7 @@ testRingpop('set adds all members', function t(deps, assert) {
 });
 
 testRingpop('set emits an event', function t(deps, assert) {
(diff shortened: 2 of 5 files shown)


test/run-shared-integration-tests
```diff
@@ -6,7 +6,7 @@ set -eo pipefail
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="master"
+declare ringpop_common_branch="fix/test-state-transistions"
```
mennopruijssers: This will be changed back to `master` after uber/ringpop-common#95 is merged into master

test/unit/membership_test.js
```diff
@@ -266,7 +266,7 @@ testRingpop('set adds all members', function t(deps, assert) {
 });
 
 testRingpop('set emits an event', function t(deps, assert) {
```
thanodnl: The name of this test doesn't cover what it is doing anymore, we should rename it, or duplicate it to only test the additional call to the `handleUpdate` function.

test/unit/state_transistion_test.js
```diff
@@ -0,0 +1,68 @@
...
+}
+
+handleUpdateSchedulesTimer(Member.Status.alive, false);
+handleUpdateSchedulesTimer(Member.Status.leave, false);
+handleUpdateSchedulesTimer(Member.Status.faulty, true);
+handleUpdateSchedulesTimer(Member.Status.suspect,true);
+handleUpdateSchedulesTimer(Member.Status.tombstone,true);
+
+function handleUpdateSchedulesTimer(state, shouldSchedule) {
+    test('handleUpdate schedules timers ', function t(assert) {
```
thanodnl: lets add the `state` value to the name for easier understanding of test output.

test/unit/state_transistion_test.js
```diff
@@ -0,0 +1,68 @@
...
+            stateTransitions.timers[address] = {state: 'fake'}
+        }
+
+        stateTransitions.handleUpdate({
+            status: state,
+            address: address
+        });
+        if (shouldSchedule) {
+            assert.ok(stateTransitions.timers[address], 'timer scheduled');
+            assert.equal(stateTransitions.timers[address].state, state, 'suspect timer scheduled');
```
thanodnl: the name here implies a specific state, it should be the correct state or a more generic line like `state transition timer scheduled`.

test/unit/state_transistion_test.js
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            assert.notok(stateTransitions.timers[address], 'timer not scheduled');
+        }
+
+
```
thanodnl: nit: double white line

severb:
LGTM, good job.

PR #314, opened 2017-01-06
Retry discover provider

mennopruijssers:
Support retrying the discover provider if it returns an error or if it does not return valid results.

### DIFF
--- a/discover-providers.js
+++ b/discover-providers.js
@@ -20,6 +20,10 @@
 
 'use strict';
 
+var _ = require('underscore');
+
+var Backoff = require('./lib/backoff');
+
 /**
  * A DiscoverProvider provides a list of peers for a ringpop.
  * @typedef {function} DiscoverProvider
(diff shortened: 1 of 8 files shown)


discover-providers.js
```diff
@@ -73,13 +77,63 @@ function createJsonFileDiscoverProvider(hostsFile) {
...
+            policy = _.defaults({}, opts, defaultPolicy);
+        }
+        var backoff = opts.backoff || new Backoff(policy);
+        var lastError = null;
+
+        attempt(null);
+
+        function attempt(fail) {
+            if (fail) {
+                return setImmediate(callback, lastError);
```
thanodnl: `lastError` might be `null` if the last call to the `innerDiscoveryProvider` didn't result in an error but in an empty host list. The result is that the `callback` that is passed to the discovery provider will return without a hosts list (`undefined` compared to an empty list) and without an error. Are we sure that the consumer of a call to the discovery provider can handle these cases?

I think it is safe to pass an empty list always to the callback to make sure it will not try to iterate over an undefined list.

test/integration/discover-provider-retry-test.js
```diff
@@ -0,0 +1,121 @@
...
+            bootstrapFile: '/tmp/does.not.exist.json',
+            retry: false
+        }, function onBootstrap(err, nodesJoined) {
+            assert.ok(err, 'error during bootstrap');
+
+            ringpop.destroy();
+            assert.end();
+        });
+});
+
```
thanodnl: nit: double white line

test/integration/discover-provider-retry-test.js
```diff
@@ -0,0 +1,121 @@
...
+        var bootstrapFileWritten = false;
+        var bootstrapFileName = tmp.tmpNameSync({});
+
+        ringpop.bootstrap({
+            bootstrapFile: bootstrapFileName,
+            retry: {
+                minDelay: 10
+            }
+        }, function onBootstrap(err, nodesJoined) {
+            assert.error(err, 'no error during bootstrap');
```
thanodnl: this reads as if it is asserting an error here, but I don't think we want an error on this path right? Is this assert the right thing here?

test/unit/discover-providers-test.js
```diff
@@ -125,5 +222,29 @@ test('discoverProvider createFromOpts', function t(assert) {
...
+
+    assert.test('opts with retry - returns retrying discover provider', function t(assert) {
+        assert.plan(5);
+
+        var first = true;
+        var fn = function exampleDiscoverProvider(cb) {
+            if(first) {
+                first = false;
+                assert.pass('inner provider called once');
+                cb(null, null);
```
thanodnl: Maybe test with an empty list as well. I think it is more realistic to expect an empty list then `null` if there are no peers.

test/integration/discover-provider-retry-test.js
mennopruijssers: according to [the docs](https://github.com/substack/tape#terrorerr-msg) `assert.error` fails if `err` is not falsy (`null`, `undefined` or `false`). I can change it to an alias (`t.ifError`) if you thinks it's easer to read?

discover-providers.js
mennopruijssers: I agree but I would prefer explicitly calling the callback with `null` as the list. `null` is falsy where an empty array is not. causing patterns like:
```
if (!hosts) {
}
```
to stop working.

mennopruijssers: After giving this another thought, it might make more sense to always pass an error to the callback. This is fairly easy by doing `lastError || fail`. What do you think?

discover-providers.js
```diff
@@ -96,27 +151,29 @@ function createFromOpts(opts) {
...
+    var discoverProvider;
     if (opts.discoverProvider) {
-        return opts.discoverProvider;
+        discoverProvider = opts.discoverProvider;
+    } else if (opts.hosts) {
+        discoverProvider = createStaticHostsProvider(opts.hosts);
+    } else if (opts.bootstrapFile && typeof opts.bootstrapFile === 'string') {
+        discoverProvider = createJsonFileDiscoverProvider(opts.bootstrapFile);
+    } else if (opts.bootstrapFile && Array.isArray(opts.bootstrapFile)) {
+        /* This weird case is added for backwards compatibility :( */
```
dansimau: nit: sad face should be happy face?

mennopruijssers: it's sad that we have a weird case :)

dansimau: As a user I think it should be happy that we have backwards compatibility 😁

mennopruijssers: true.

PR #315, opened 2017-01-19
[Fix] Consistent hash ring lookups 1 / Use comparator function

mennopruijssers:
This is the first part of the fix for consistent hashring lookups on hash collisions where we abstract the RBTree to not care about the datatypes stored. Comparison is done by a comparator function.

This will help in later stages where the `key` can contain more data points instead of the hash.

### DIFF
--- a/test/unit/rbtree_test.js
+++ b/test/unit/rbtree_test.js
@@ -68,19 +68,19 @@ function validateRBTree(node) {
     var rightNode = node.right;
 
     if (isRed(node) && (isRed(leftNode) || isRed(rightNode))) {
-        throw new Error('red violation at node val' + node.val);
+        throw new Error('red violation at node key' + node.key);
     }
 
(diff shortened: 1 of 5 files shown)


test/unit/rbtree_test.js
```diff
@@ -20,10 +20,11 @@
 
 var RBTree = require('../../lib/ring/rbtree').RBTree;
 var RingNode = require('../../lib/ring/rbtree').RingNode;
+var comparator = require('../../lib/ring').comparator;
```
thanodnl: I think it is best to define the comparator function in the test instead of reusing the one in the code so we don't have to change the tests when we change the key/value in the RBTree for the HashRing.

mennopruijssers: The next PR does exactly this. For this PR, I prefer to re-use the comparator from `ring/index.js` instead of duplicating the same function in the tests

PR #316, opened 2017-01-20
[Fix] Consistent hash ring lookups 2 / Use replicaPoints as RBTree key

mennopruijssers:
This PR changes the key in the RBTree from the hash to a replica point in preparation for a better compare function.

### DIFF
--- a/lib/ring/index.js
+++ b/lib/ring/index.js
@@ -34,8 +34,9 @@ function HashRing(options) {
     this.servers = {};
     this.checksum = null;
 }
+
 HashRing.comparator = function comparator(a, b) {
-    return a - b;
+    return a.hash - b.hash;
 };
 
 util.inherits(HashRing, EventEmitter);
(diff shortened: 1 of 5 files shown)


lib/ring/index.js
```diff
@@ -34,8 +34,9 @@ function HashRing(options) {
     this.servers = {};
     this.checksum = null;
 }
+
 HashRing.comparator = function comparator(a, b) {
-    return a - b;
+    return a.hash - b.hash;
```
mennopruijssers: This needs a test

mennopruijssers: done

test/unit/hashring_test.js
```diff
@@ -165,3 +165,16 @@ test('servers removed out of order result in same checksum', function t(assert)
         }
     }
 });
+
+test('hashring replica point comparator', function t(assert) {
```
mennopruijssers: These tests seems silly now but will be extended when the comparator uses more than just the hash for comparison.

PR #317, opened 2017-01-20
[Fix] Consistent hash ring lookups 3 / consistent lookups on hash collisions

mennopruijssers:
With the two previous PR's in place, the stage is set for the actual fix. By comparing not just the hash but also the addresses of two replica points in the hash ring, we know for sure that the order  is guaranteed -- even when there are hash collisions.

### DIFF
--- a/test/unit/hashring_test.js
+++ b/test/unit/hashring_test.js
@@ -166,6 +166,55 @@ test('servers removed out of order result in same checksum', function t(assert) 
     }
 });
 
+test('hashring consistent lookups on collision - synthetic collision', function t(assert) {
+    var ring1 = new HashRing();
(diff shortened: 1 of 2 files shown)


test/unit/hashring_test.js
```diff
@@ -166,6 +166,57 @@ test('servers removed out of order result in same checksum', function t(assert)
     }
 });
 
+test('hashring consistent lookups on collision - synthetic collision', function t(assert) {
+    var ring1 = new HashRing();
+    var ring2 = new HashRing();
+
+    // These 2 host:ports will cause a hash collusion due to the concat of
```
severb: s/collusion/collision/

PR #319, opened 2017-01-27
Fix licence header after uber-licence update

mennopruijssers:
`npm run add-licence` after clean `npm install`

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "timer-shim": "^0.3.0",
     "tmp": "0.0.31",
     "tryit": "^1.0.1",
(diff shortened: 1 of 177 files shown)


package.json
```diff
@@ -68,7 +68,7 @@
     "timer-shim": "^0.3.0",
     "tmp": "0.0.31",
     "tryit": "^1.0.1",
-    "uber-licence": "^2.0.1",
+    "uber-licence": "^2.1.3",
```
mennopruijssers: This is to make sure all the devs (tm) are updated after running `npm install`

PR #320, opened 2017-04-04
Allow clients to customize ringpop stats prefix

yulunli:


mennopruijssers:
Overall looks good. Could you add tests to make sure the option is propagated correctly? Also, It looks like `statHostPort` is never used (except for one test). Lets remove it?

index.js
```diff
@@ -195,9 +195,13 @@ function RingPop(options) {
     this.serverRate = new metrics.Meter();
     this.totalRate = new metrics.Meter();
 
-    // 10.30.8.26:20600 -> 10_30_8_26_20600
-    this.statHostPort = this.hostPort.replace(/\.|:/g, '_');
-    this.statPrefix = 'ringpop.' + this.statHostPort;
+    if (options.statPrefix) {
```
mennopruijssers: Maybe explicitly check for `undefined`. So `false` can be used to NOT emit a stat prefix

mennopruijssers: After offline discussion we decided to not allow this.

PR #323, opened 2017-06-15
Add option to enforce identical app name in pings

yulunli:
Always reject pings from wrong app name

### DIFF
--- a/server/protocol/ping.js
+++ b/server/protocol/ping.js
@@ -40,6 +40,18 @@ module.exports = function createPingHandler(ringpop) {
             return callback(new Error('need req body with source, changes, and checksum'));
         }
 
+        if (body.hasOwnProperty('app')) {
+            if (body.app !== ringpop.app) {
+                ringpop.logger.warn('Rejected ping from wrong ringpop app', body);
+                return callback(new Error('Pinged ringpop has a different app name'));
+            }
+        } else {
+            if (ringpop.requiresAppInPing) {
+                ringpop.logger.warn('Rejected ping from unknown ringpop app', body);
+                return callback(new Error('Pinged ringpop requires app name'));
+            }
+        }
+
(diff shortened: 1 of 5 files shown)


server/protocol/ping.js
```diff
@@ -40,6 +40,18 @@ module.exports = function createPingHandler(ringpop) {
             return callback(new Error('need req body with source, changes, and checksum'));
         }
 
+        if (body.hasOwnProperty('app')) {
+            if (body.app !== ringpop.app) {
+                ringpop.logger.warn('Rejected ping from wrong ringpop app', body);
```
mennopruijssers: log as error?

server/protocol/ping.js
```diff
@@ -40,6 +40,18 @@ module.exports = function createPingHandler(ringpop) {
...
+                return callback(new Error('Pinged ringpop has a different app name'));
+            }
+        } else {
+            if (ringpop.requiresAppInPing) {
+                ringpop.logger.warn('Rejected ping from unknown ringpop app', body);
```
mennopruijssers: same?

mennopruijssers:
Oh, I just saw ringpop-common#97. Can you add the integration test to that one and temporary change the branch name in [run-shared-integration-tests:9](https://github.com/uber-node/ringpop-node/blob/master/test/run-shared-integration-tests#L9) temporary until ringpop-common#97 is merged

server/protocol/ping.js
yulunli: I think warning is better because if another service is pinging incorrectly, it's not an error for the pinged service.

proflayton:
this lgtm

yulunli:
Given we already have a test in ringpop-node, the test in ringpop-common will be added at a later time.

